PR #49, opened 2015-02-11
Don't call UpdateGfxMenu (gone alongside lilo support)

ancorgs:


jsuchome:
Hm, I thought it's already gone. +1

jreidinger:
LGTM, I think it is just not backported

ancorgs:
The same line is still in master, I'll merge the fix now

PR #51, opened 2015-02-13
Vmware support (bnc#773323)

lslezak:
On VMware we need to propose local time, see https://bugzilla.suse.com/show_bug.cgi?id=773323

- Unified RSpec tests (via a shared `test/test_helper.rb`)
- Enabled code coverage + coveralls


timezone/src/clients/timezone_proposal.rb
```diff
@@ -28,15 +26,7 @@ def main
...
-        @m = Convert.to_map(
-          SCR.Execute(path(".target.bash_output"), "/bin/date +%Y")
-        )
-        Builtins.y2milestone("date call: %1", @m)
-
-        if Ops.less_than(
-            Builtins.tointeger(Ops.get_string(@m, "stdout", "0")),
-            2004
-          )
+        if Time.now.year < 2015
```
ancorgs: It's already better than before, but in the future we should probably do something more... well... elegant (like using the creation/modification time of some file) in order to decide if we are in the past.

timezone/test/Timezone_test.rb
```diff
@@ -0,0 +1,46 @@
...
+Yast.import "Timezone"
+
+describe Yast::Timezone do
+  subject { Yast::Timezone }
+
+  describe "#ProposeLocaltime" do
+    it "returns true if a Windows partition is found" do
+      subject.windows_partition = true
+      expect(Yast::SCR).to receive(:Read)
+        .with(Yast::Path.new(".probe.is_vmware")).and_return(false)
```
ancorgs: in ```yast/rspec``` we have a ```path``` shortcut.

ancorgs: NP: I'd use ```allow``` instead of ```expect``` to mock the the SCR call. Just a semantic puristic issue :-)

timezone/test/Timezone_test.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+
+Yast.import "Timezone"
+
+describe Yast::Timezone do
+  subject { Yast::Timezone }
```
ancorgs: I would use ```subject { Yast::Timezone.ProposeLocalTime }``` inside the ```describe``` below, instead of this ```subject``` here.

timezone/src/clients/timezone_proposal.rb
lslezak: Yes, using a fixed date does not work well, this hasn't been updated since 2004...
I'll fix that.

ancorgs:
Sorry to turn this into a neverending story, but now that you are using ```yast/rspec``` the ```SCRStub``` (actually a precursor of the classes at ```yast/rspec```) and the tests using it should be adjusted. I can do it myself adding some commit to this PR, if you wish (it's shorter than explaining it).

ancorgs:
Updated. It actually took me a while to re-discover the real meaning of the workaround for bnc#891053. BTW, we should add a dependencies for yast-core including the fix and ruby-binding including yast/rspec. I'll do now

ancorgs:
Update. BTW @lslezak, did you try the mtime thingy manually? We should probably have a test for it, anyways, although I don't consider it mandatory for the merge (I would say the manual test is).

lslezak:
Yes, I manually tested the `mtime` change in `irb` and I checked that the mtime date of the file in inst-sys is reasonable (e.g. not 1.1.1970 or something like that).

lslezak:
@ancorgs thanks for the test improvements and dependency update!

ancorgs:
Then, if it looks good to you, it looks good to me. :-)

lslezak:
OK, mutual LGTM :wink:

PR #52, opened 2015-02-20
Alert the user of errors (bnc#918093)

ancorgs:
Some zombie killing and test refactoring in the way.

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -1038,6 +1038,10 @@ module Yast
           log.info "Not setting X keyboard due to text mode"
         # check if we are running over ssh: bnc#539218,c4
         elsif x11_over_ssh?
+          # TODO: the check above could not be enough in some cases
+          # An external X server can be specified via display_ip boot parameter
+          # (see https://en.opensuse.org/SDB:Linuxrc#p_displayip).
+          # I that case, the configuration should probably also be skipped
           log.info "Not setting X keyboard: running over ssh"
         elsif !@xkb_cmd.empty?
           SetKeyboard(keyboard)
@@ -1490,7 +1494,7 @@ module Yast
     display.split(":")[1].to_i >= 10
   end
 
-  # Checks if it's running is text mode (no X11)
+  # Checks if it's running in text mode (no X11)
   def textmode?
     if !Stage.initial || Mode.live_installation
       UI.TextMode
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -120,7 +120,7 @@ module Yast
       Yast.import "ProductFeatures"
       Yast.import "Stage"
       Yast.import "XVersion"
-      Yast.import "Popup"
+      Yast.import "Report"
 
       # ------------------------------------------------------------------------
       # START: Globally defined data to be accessed via Keyboard::<variable>
@@ -1486,29 +1486,27 @@ module Yast
   # Checks if the graphical environment is being executed remotely using
   # "ssh -X"
   def x11_over_ssh?
-    display = Builtins.getenv("DISPLAY") || ""
+    display = ENV["DISPLAY"] || ""
     display.split(":")[1].to_i >= 10
   end
 
   # Checks if it's running is text mode (no X11)
   def textmode?
-    textmode = Linuxrc.text
     if !Stage.initial || Mode.live_installation
-      display_info = UI.GetDisplayInfo
-      textmode = display_info && display_info["TextMode"]
+      UI.TextMode
+    else
+      Linuxrc.text
     end
-    # Ensure it's boolean
-    !!textmode
   end
 
   # Executes the command to set the keyboard in X11, reporting
   # any error to the user
   def execute_xkb_cmd
     log.info "Setting X11 keyboard to: <#{@current_kbd}>"
-    log.info "Setting X11 keyboard:\n #{@xkb_cmd}"
+    log.info "Setting X11 keyboard: #{@xkb_cmd}"
     if SCR.Execute(path(".target.bash"), @xkb_cmd) != 0
       log.error "Failed to execute the command"
-      Popup::Error(_("Failed to set X11 keyboard to '%s'") % @current_kbd)
+      Report::Error(_("Failed to set X11 keyboard to '%s'") % @current_kbd)
     end
   end
 
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -120,6 +120,7 @@ module Yast
       Yast.import "ProductFeatures"
       Yast.import "Stage"
       Yast.import "XVersion"
+      Yast.import "Popup"
 
       # ------------------------------------------------------------------------
       # START: Globally defined data to be accessed via Keyboard::<variable>
@@ -1021,7 +1022,6 @@ module Yast
     end # SetConsole()
 
 
-
     # Set the X11 keyboard to the given keyboard language.
     #
     # @param	Keyboard language e.g.  "english-us"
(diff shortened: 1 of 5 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
     end
   end
 
+  # Checks if the graphical environment is being executed remotely using
+  # "ssh -X"
+  def x11_over_ssh?
+    display = Builtins.getenv("DISPLAY") || ""
```
lslezak: NP: One more zombie to kill: `Builtins.getenv("DISPLAY")` => `ENV["DISPLAY"]`

keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
...
+    display.split(":")[1].to_i >= 10
+  end
+
+  # Checks if it's running is text mode (no X11)
+  def textmode?
+    textmode = Linuxrc.text
+    if !Stage.initial || Mode.live_installation
+      display_info = UI.GetDisplayInfo
```
lslezak: You can use `UI.TextMode` directly (returns boolean).

keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
...
+    end
+    # Ensure it's boolean
+    !!textmode
+  end
+
+  # Executes the command to set the keyboard in X11, reporting
+  # any error to the user
+  def execute_xkb_cmd
+    log.info "Setting X11 keyboard to: <#{@current_kbd}>"
+    log.info "Setting X11 keyboard:\n #{@xkb_cmd}"
```
lslezak: Avoid newlines in log messages, that breaks log formatting. Long lines are OK in the log...

keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
     end
   end
 
+  # Checks if the graphical environment is being executed remotely using
+  # "ssh -X"
+  def x11_over_ssh?
+    display = Builtins.getenv("DISPLAY") || ""
+    display.split(":")[1].to_i >= 10
```
lslezak: What about `display[1..-1].to_i` ? AFAIK when using forwarded X there is no hostname in `DISPLAY` in front of `:`...

kobliha: IMO display.split(":")[1].to_i is a bit more intuitive than display[1..-1].to_i

Using ENV["DISPLAY"] is a good idea: ENV["DISPLAY"].split(":")[1].to_i

kobliha: BTW, `10` sounds like a magic constant

keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
...
+  end
+
+  # Executes the command to set the keyboard in X11, reporting
+  # any error to the user
+  def execute_xkb_cmd
+    log.info "Setting X11 keyboard to: <#{@current_kbd}>"
+    log.info "Setting X11 keyboard:\n #{@xkb_cmd}"
+    if SCR.Execute(path(".target.bash"), @xkb_cmd) != 0
+      log.error "Failed to execute the command"
+      Popup::Error(_("Failed to set X11 keyboard to '%s'") % @current_kbd)
```
kobliha: In case this could run in AutoYast, Report.Error should be used if I'm not mistaken

keyboard/src/modules/Keyboard.rb
ancorgs: @lslezak even if you are right and there is no hostname, changing the way we do the check sounds risky with no clear benefit. I'll change the call to ```getenv``` but keep the old approach.

keyboard/src/modules/Keyboard.rb
```diff
@@ -1030,40 +1030,20 @@ def SetConsole(keyboard)
...
         # check if we are running over ssh: bnc#539218,c4
-        elsif Ops.greater_or_equal(
-            Builtins.tointeger(
-              Ops.get(Builtins.splitstring(display, ":"), 1, "0")
-            ),
-            10
-          )
-          Builtins.y2milestone("Not setting X keyboard: running over ssh")
-        elsif Ops.greater_than(Builtins.size(@xkb_cmd), 0)
+        elsif x11_over_ssh?
```
lslezak: JFYI: you can specify any external X server to use for installation via `display_ip` boot parameter (see https://en.opensuse.org/SDB:Linuxrc#p_displayip). I guess in that case we should also skip setting keyboard.

In the end we should probably set the keyboard only when `DISPLAY` is set to `:0`...

keyboard/src/modules/Keyboard.rb
```diff
@@ -1504,6 +1483,43 @@ def write_udev_rule
...
+      log.error "Failed to execute the command"
+      Popup::Error(_("Failed to set X11 keyboard to '%s'") % @current_kbd)
+    end
+  end
+
+  # Enables autorepeat if needed
+  def enable_autorepeat
+    return nil unless Stage.initial && !Mode.live_installation && !xen_running
+    cmd = "xset r on"
+    log.info "calling xset to fix autorepeat problem: #{cmd}"
```
kobliha: You have forgotten to mention: `bnc#371756: enable autorepeat`

Nobody would know which autorepeat problem it was in the past and thus would have no idea if that is still needed at all

keyboard/src/modules/Keyboard.rb
lslezak: OK, I'm fine with that...

keyboard/src/modules/Keyboard.rb
ancorgs: It's in the call to the method, to be coherent with the call to ```write_udev_rule``` in which the comment about the bug is also in the caller.

kobliha: Ah, yep, you're right

keyboard/src/modules/Keyboard.rb
ancorgs: Yeah, actually methods like ```textmode?``` or ```x11_over_ssh?``` is the kind of things that you would not expect to have to implement in every module/client. Although I suspect that all the implementations of similar checks we have spread all over the YaST code have slightly different semantic from each other, so not so easy to unify.

keyboard/src/modules/Keyboard.rb
ancorgs: I was unsure if it was exactly the same. I'll kill him!

keyboard/src/modules/Keyboard.rb
ancorgs: I suspect that 10 comes from the default value of X11DisplayOffset. You know, one of those things that _almost_ nobody changes.
http://becausewecan.org/node/65

But my lack of understanding on X11 over ssh and all the weird scenarios that resulted in somebody filling bnc#539218, makes me be conservative here :-)

keyboard/src/modules/Keyboard.rb
kobliha: And IMO we should do something about it, Trello card, Bugzilla report, Issue - just anything, otherwise this will get lost

kobliha:
LGTM :) @lslezak ?

lslezak:
LGTM (But we should notice https://github.com/yast/yast-country/pull/52#discussion_r25080841 for later if not fixing it now.)

ancorgs:
I added a comment about display_ip and fixed the typo in the other comment. I'll merge as soon as travis passes.

keyboard/src/modules/Keyboard.rb
ancorgs: I created
https://trello.com/c/qSzdaYn3/283-remove-redundant-checks-spread-over-several-modules

PR #53, opened 2015-06-23
Use Ho Chi Minh City in UI and proposals (bnc#923506)

ancorgs:


jreidinger:
Well, it is politically sensible question: http://goasia.about.com/od/Language/fl/Is-It-Ho-Chi-Minh-City-or-Saigon.htm
So I am not sure if it is so straightforward, but as it is official name, then LGTM

ancorgs:
Well, the actual internal name of the timezone was already changed in YaST in 2008 and I fear the proposal has been broken since then. Let's unify!

PR #54, opened 2015-07-24
Dependency cleanup

mvidner:
actually that is just the user visible change which is not that important.
The important thing is "Coverage report: do not count tests." which I want in a separate PR so that my *next* PR has an *increase* in test coverage :)


ancorgs:
LGTM

PR #55, opened 2015-08-04
Ensure ntp is installed (when Install Recommended Packages is off)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=938951

The actual fix is https://github.com/yast/yast-country/commit/13e61a9e8d5f60d8ecacb301b4429a36e1c7a2c3 and the rest is refactoring. But the code is spread across two packages (see a related PR in ntp-client) and I am not sure how much of a cleanup I should attempt now (and even if this is in the end a cleanup)

### DIFF
--- a/timezone/src/lib/timezone/ntp.rb
+++ b/timezone/src/lib/timezone/ntp.rb
@@ -1,35 +1,47 @@
 
 module Yast
-  # Acts as interface of the Timezone package to NTP functionality
-
   # TODO: separate UI and non-UI, figure out a reasonable UI interface
   # (check also CWM)
+
+  # Acts as interface of the Timezone package to NTP functionality
   class TimezoneNtp
     include Yast::UIShortcuts
 
-    # [Boolean] if system clock is configured to sync with NTP
-    attr_accessor :used
+    # @return [Boolean] if system clock is configured to sync with NTP
+    attr_accessor :used         # (Writing it also notifies NtpClient)
 
-    # [String] ntp server configured to sync with
-    # FIXME we use an array, but this is the one to contact now
+    def used=(value)
+      @used = value
+      ntp_call("SetUseNTP", { "ntp_used" => value })
+    end
+
+    # @return [Boolean] used in the installed system
+    def enabled?
+      ntp_call("GetNTPEnabled", {}) == true
+    end
+
+    # @return [String] ntp server configured to sync with
     attr_accessor :server
 
     def initialize
-      self.used = false
-      self.server = ""          # FIXME better nil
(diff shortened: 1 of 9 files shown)


timezone/src/lib/timezone/ntp.rb
```diff
@@ -0,0 +1,174 @@
...
+
+    # This proxy handles the complication
+    # that the package yast2-ntp-client may not be present
+    # (on the running system).
+    # The called client has been meanwhile moved to this package but it
+    # does not check for the dependencies itself yet.
+    def ntp_call(acall, args)
+      if !@installed
+        # replace "replace_point" by the widgets
+        if acall == "ui_init"
```
lslezak: NP: I know this was simply copied from elsewhere, but using `case...when` here would really improve it a lot...

timezone/src/lib/timezone/ntp_adapter.rb
```diff
@@ -0,0 +1,476 @@
...
+        @func,
+        @param
+      )
+
+      # FIXME must go to module to preserve value
+      @ntp_was_used = false
+
+
+
+      if false
```
lslezak: NP: This `if false` is useless, and the rest could be replaced by `case...when` as well...

timezone/src/lib/timezone/ntp_adapter.rb
```diff
@@ -0,0 +1,476 @@
...
+              required_package
+            )
+          )
+        end
+      end
+
+      ret = 0
+      if NetworkService.isNetworkRunning
+        # Only if network is running try to synchronize the ntp server
+        Popup.ShowFeedback("", _("Synchronizing with NTP server..."))
```
lslezak: NP: The new `Popup.Feedback` with a block should be used here...

ancorgs:
To be honest, such a big refactoring (although a big part is just code moving) without unit test scares me a bit.

mvidner:
@ancorgs You're right, and the API redesign is not too good either; so I am going to improve that and add tests, but proceed with the actual simple fix in another PR.

jreidinger:
@mvidner what is status here? I worry it is too risky for SP1 now, but if you improve it a bit, it can be merged for SP2

kobliha:
@mvidner SP2 is just behind the door, any progress?

teclator:
There has not been updates since a while and the bug was already fixed by https://github.com/yast/yast-ntp-client/pull/34 so maybe make sense to close it by now and just create a new PR if some improvements are done at the end.

@mvidner , what do you think about?

PR #56, opened 2015-08-14
- use xkb converted console keymaps where applicable (FATE#318426)

sndirsch:


ancorgs:
Right now, we are sharing the master branch of the YaST repositories for both SLE12-SP1 and Factory. So I need to branch SP1 before merging this. Discussing it right now within the YaST team.

ancorgs:
Replaced by #57

PR #57, opened 2015-08-17
use xkb converted console keymaps where applicable (FATE#318426)

ancorgs:
Replaces #56 adding fixed unit tests.


jreidinger:
LGTM

PR #59, opened 2015-10-19
SLE12-SP1 & Leap 42.1: Allocate space for "Other Settings" button proportionally to the translation lengt

kobliha:
* bsc#946955
* New screenshots are at https://trello.com/c/OrH3F8uW/348-2-sle12-sp1-p1-946955-yast-date-and-time-not-enough-space-for-button
* Code already LGTMed, this is a backport to SLE12-SP1 & Leap 42.1 (reported as different bug)

PR #60, opened 2015-10-19
Mergin SP1 changes to master

kobliha:
According to http://yastgithubio.readthedocs.org/en/latest/maintenance-branches/ and PR #59 we need to merge back to master even if no changes are actually done. I also needed to remove Rakefile changes done for SLE 12 SP1 / Leap 42.1.


ancorgs:
lgtm

PR #72, opened 2016-07-07
Use the new official short name for the Czech Republic: Czechia

jiribohac:
The short name is now officially entered in the UNGEGN database
as of 2016-07-05: http://unstats.un.org/unsd/geoinfo/geonames/

### DIFF
--- a/timezone/src/data/timezone_raw.ycp
+++ b/timezone/src/data/timezone_raw.ycp
@@ -65,7 +65,7 @@ $[
 	"Europe/Paris" : _("France"),
 	// time zone
 	"Europe/Podgorica"	: _("Montenegro"),
-	"Europe/Prague" : _("Czech Republic"),
+	"Europe/Prague" : _("Czechia"),
 	"Europe/Riga" : _("Latvia"),
 	"Europe/Rome" : _("Italy"),
 	// time zone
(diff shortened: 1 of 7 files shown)


jsuchome:
Well, not that I care that much, but this place does not look like the one that really needs the short form...

jiribohac:
On Mon, Jul 11, 2016 at 02:55:32AM -0700, Jiří Suchomel wrote:
> Well, not that I care that much, but this place does not look like the one that really needs the short form...

have a look at the list.

It says "France", not "French Republic"
It says "Germany", not "Federal Republic of Germany"
So it should say "Czechia", not "Czech Republic"

:)

-- 
Jiri Bohac
e-mail/jabber: jiri@boha.cz

jsuchome:
Hm, it's true that the list uses short forms whenever possible...

jreidinger:
@jsrain @kobliha as it is a bit controversial change, please decide. Technically it is correct change.

kobliha:
Let's stick to what we have been using for ages, `Czech Republic`. I don't see any reason for changing it here an now.

kobliha:
Thanks for your PR. We haven't found a good reason for the change though, so closing.

PR #73, opened 2016-07-10
Explanation for hwclock setting

sebix:
Reading the help does not clarify possible values. I thought the value should be a time.


jreidinger:
looks good, thanks for patch.

PR #76, opened 2016-07-14
Configure rake to commit to YaST:storage-ng

ancorgs:
This goes together with https://ci.opensuse.org/view/Yast/job/yast-country-storage-ng/


jreidinger:
LGTM

PR #77, opened 2016-07-25
Asturian keyboard support

ancorgs:


jreidinger:
LGTM, we should get rid of that ycp data files

mvidner:
@ancorgs Does this depend on a feature added only recently to an underlying package? If yes, then we should note it in the changelog. (A RPM dependency is not appropriate)

ancorgs:
@mvidner I believe support for Asturian has been in the underlying system since we adopted systemd. It was only missing in YaST, AFAIK.

mvidner:
OK, somehow I got the impression that we were waiting for an updated package. If not, fine.
LGTM

PR #61, opened 2015-10-28
Some changes in the tumbleweed mappings (bsc#942899 and bsc#942896)

ancorgs:

PR #78, opened 2016-08-10
Changelog entry missing in pull request #61

ancorgs:
https://github.com/yast/yast-country/pull/61 was merged without a changelog entry and that caused some problems to Factory maintainers. This should fix that (I will submit the fix manually if Factory maintainers want me to).


wfeldt:
lgtm

kobliha:
Also LGTM, let's just try :)

PR #79, opened 2016-08-10
Added check for supported languages and correcting unsupported ones

kobliha:
### Added check for importing invalid language (AutoYast)
- bnc#991001
- AutoYast did not report erroneous `<language/>` and used anything that came from the profile. That was causing problems, e.g., with systemctl output
- This patch adds a check for valid `<language/>` and if it's incorrect, error is reported and default language is used instead
- A little refactoring needed to be done to increase the readability of the code and also to make it testable with unit tests

autoyast - check for incorrect language

### Added new test for Yast::Language
- Some pieces might look a bit crazy, but they are caused by the fact that Language uses a constructor

### What is not changed
- Refactoring was indeed limited, not everything that deserves fixing/refactoring was changed, but the code is better
- For instance, `adapt_install_inf` is a new function, but contains only a moved code, without refactoring
- Further refactoring was needed

### What's next
- We need to port this to SLE 12 SP2, but that needs to be done after this PR is merged

### DIFF
--- a/language/src/modules/Language.rb
+++ b/language/src/modules/Language.rb
@@ -403,17 +403,16 @@ module Yast
       # TRANSLATORS: Error message. Strings marked %{...} will be replaced
       # with variable content - do not translate them, please.
       Report.Error(
-        _("Language '%{language}' was not found within the list of supported languages
-available at %{directory}.
-
-Fallback language %{fallback} will be used." % {
+        _("Language '%{language}' was not found within the list of supported languages\n" +
+          "available at %{directory}.\n\nFallback language %{fallback} will be used."
+        ) % {
           :language => language,
           :directory => @languages_directory,
           :fallback => DEFAULT_FALLBACK_LANGUAGE
-        })
+        }
       )
 
-      language.replace(DEFAULT_FALLBACK_LANGUAGE)
+      return DEFAULT_FALLBACK_LANGUAGE
     end
 
     # Changes the install.inf in inst-sys according to newly selected language
@@ -462,7 +461,7 @@ Fallback language %{fallback} will be used." % {
       )
 
       if @language != lang
-        correct_language(lang)
+        lang = correct_language(lang)
 
         if Stage.initial && !Mode.test && !Mode.live_installation
           integrate_inst_sys_extension(@language)
@@ -474,7 +473,7 @@ Fallback language %{fallback} will be used." % {
         # In config mode, use language name translated into the current language
         # othewrwise use the language name translated into that selected language
         # because the whole UI will get translated later too
-        @name = Mode.config ? language_def[4] : language_def[0] || lang
+        @name = (Mode.config ? language_def[4] : language_def[0]) || lang
         @language = lang
         Encoding.SetEncLang(@language)
       end
--- a/language/src/modules/Language.rb
+++ b/language/src/modules/Language.rb
@@ -31,6 +31,10 @@ require "yast"
 
 module Yast
   class LanguageClass < Module
+    DEFAULT_FALLBACK_LANGUAGE = "en_US".freeze
+
+    include Yast::Logger
+
     def main
       Yast.import "Pkg"
       Yast.import "UI"
@@ -49,15 +53,19 @@ module Yast
       Yast.import "PackageSystem"
       Yast.import "Popup"
       Yast.import "ProductFeatures"
+      Yast.import "Report"
       Yast.import "SlideShow"
       Yast.import "Stage"
 
+      # directory where all the language definitions are stored
+      # it's a constant, but depends on a dynamic content (Directory)
+      @languages_directory = "#{Directory.datadir}/languages"
 
       # currently selected language
-      @language = "en_US"
+      @language = DEFAULT_FALLBACK_LANGUAGE
 
       # original language
-      @language_on_entry = "en_US"
+      @language_on_entry = DEFAULT_FALLBACK_LANGUAGE
 
       # language preselected in /etc/install.inf
       @preselected = "en_US"
@@ -68,12 +76,11 @@ module Yast
       @linuxrc_language_set = false
 
       # Default language to be restored with MakeProposal.
-      @default_language = "en_US"
+      @default_language = DEFAULT_FALLBACK_LANGUAGE
 
 
       # Default settings for ROOT_USES_LANG in /etc/sysconfig/language
       @rootlang = "ctype"
-
 
       # Default settings for INSTALLED_LANGUAGES in /etc/sysconfig/language
       @languages = ""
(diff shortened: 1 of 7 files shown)


language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
+    # @see @languages_directory
+    def valid_language?(language)
+      GetLanguagesMap(false).key?(language)
+    end
+
+    # Checks whether given language is supported by the installer
+    # and changes it to the default language en_US if it isn't.
+    #
+    # @param [String] reference to the new language
+    # @return [String] new (corrected) language
```
lslezak: Um, "new" is a bit misleading, it's not a new instance, it's the original object with changed content.

But the API looks strange, either change the argument in place and do not return anything or return a new object. Returning the original argument does not make sense.

And I'd prefer returning a new object in this case, it would avoid that added `deep_copy` call.

language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
+          :fallback => DEFAULT_FALLBACK_LANGUAGE
+        })
+      )
+
+      language.replace(DEFAULT_FALLBACK_LANGUAGE)
+    end
+
+    # Changes the install.inf in inst-sys according to newly selected language
+    #
+    # FIXME: code just moved, refactoring needed
```
lslezak: Oh, I was about to complain about ugly code but then I noticed this `FIXME`... :wink:

language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
 
         GetLocales() if Builtins.size(@locales) == 0
 
-        @name = Ops.get_string(@languages_map, [lang, 0], lang)
-        @name = Ops.get_string(@languages_map, [lang, 4], lang) if Mode.config
+        language_def = GetLanguagesMap(false).fetch(lang, [])
+        # In config mode, use language name translated into the current language
+        # othewrwise use the language name translated into that selected language
+        # because the whole UI will get translated later too
+        @name = Mode.config ? language_def[4] : language_def[0] || lang
```
lslezak: Um, I'd write parenthesis here to make it more readable, I'd rather not rely on operator precedence here...

And what do you actually want to use here?

```ruby
Mode.config ? language_def[4] : (language_def[0] || lang)
```
or
```ruby
(Mode.config ? language_def[4] : language_def[0]) || lang
```

language/test/Language_test.rb
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+
+Yast.import "Mode"
+Yast::Mode.SetMode("normal")
```
lslezak: This should not be needed, you call `allow(Yast::Mode).to receive(:mode).and_return("normal")` in a test below.

language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
+    # including giving a UI feedback to the user
+    #
+    # @param [String] language, e.g. 'de_DE'
+    def integrate_inst_sys_extension(language)
+      log.info "integrating translation extension..."
+
+      # busy message
+      Popup.ShowFeedback(
+        "",
+        _("Downloading installation system language extension...")
```
lslezak: NP: Sounds too long to me, might not fit or look nice in 80x25 terminal... Think about some extra long translations! Moreover the popup can close before you finish reading.

Simple `Downloading language extension...` is enough IMHO.

language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
+      # No correction needed, this is already a correct language definition
+      return language if valid_language?(language)
+
+      # TRANSLATORS: Error message. Strings marked %{...} will be replaced
+      # with variable content - do not translate them, please.
+      Report.Error(
+        _("Language '%{language}' was not found within the list of supported languages
+available at %{directory}.
+
+Fallback language %{fallback} will be used." % {
```
lslezak: The variable interpolation should happen *after* translating the text otherwise the translation won't be found.
I.e. `_("foo") % bar` instead of ~~`_("foo" % bar)`~~

language/src/modules/Language.rb
```diff
@@ -363,70 +362,129 @@ def GetLanguageExtensionFilename(language)
...
         "original language: %1; setting to lang:%2",
         @language,
         lang
       )
 
-      if @language != lang # Do it only if different
+      if @language != lang
+        correct_language(lang)
+
         if Stage.initial && !Mode.test && !Mode.live_installation
```
lslezak: NP: Kill this `Mode.test` if it is not needed.

language/src/modules/Language.rb
kobliha: This, again, is just moved code... including the message

language/src/modules/Language.rb
kobliha: OK, thanks for noticing

language/src/modules/Language.rb
kobliha: I'm afraid it is, we still have the old testsuite here

language/src/modules/Language.rb
kobliha: OK, fine

language/src/modules/Language.rb
lslezak: Oh, it was just moved, so keep it.

language/test/Language_test.rb
```diff
@@ -0,0 +1,146 @@
...
+      it "does nothing with the given language" do
+        allow(subject).to receive(:valid_language?).with("known_language").and_return(true)
+
+        language = subject.correct_language("known_language")
+        expect(language).to eq("known_language")
+      end
+    end
+
+    context "when called with an unknown language" do
+      it "reports an error and adjusts the given language to a default one" do
```
lslezak: Now it returns the default language.

language/test/Language_test.rb
```diff
@@ -0,0 +1,146 @@
...
+        expect(subject).not_to receive(:adapt_install_inf)
+
+        subject.Set(new_language)
+        expect(subject.GetName).to eq(translated_language_name)
+      end
+    end
+
+    # This is a special case when we start installer in non-CJK language and then switch
+    # to a CJK one (CJK == Chinese, Japanese, and Korean), in that case, needed fonts are
+    # not loaded and the UI just cna't display these CJK characters
```
lslezak: Typo `can't`.

lslezak:
LGTM :exclamation: :tada:

kobliha:
@lslezak Thanks ;)

PR #89, opened 2016-09-23
- readded check for Windows based on storage-ng

aschnell:
For https://trello.com/c/YTeaoJYo/411-3-storageng-do-real-check-for-the-windows-partition.

### DIFF
--- a/timezone/src/clients/inst_timezone.rb
+++ b/timezone/src/clients/inst_timezone.rb
@@ -21,9 +21,6 @@
 # Client for initial timezone setting (part of installation sequence)
 # Author:	Jiri Suchomel <jsuchome@suse.cz>
 
-require "storage"
-require "y2storage"
-
 module Yast
   class InstTimezoneClient < Client
     def main
@@ -31,33 +28,19 @@ module Yast
       Yast.import "GetInstArgs"
       Yast.import "Stage"
       Yast.import "Wizard"
+      Yast.import "Timezone"
 
       Yast.include self, "timezone/dialogs.rb"
 
       @args = GetInstArgs.argmap
       @args["first_run"] = "yes" unless @args["first_run"] == "no"
 
-      if Stage.initial && system_has_windows?
+      if Stage.initial && Timezone.system_has_windows?
         Timezone.windows_partition = true
         Builtins.y2milestone("windows partition found: assuming local time")
       end
 
       full_size_timezone_dialog
-    end
-
-    # Checks whether the system has Windows installed on a primary partition with NTFS
-    def system_has_windows?
-      storage = Y2Storage::StorageManager.instance
-      Storage::Ntfs::all(storage.probed).each do |ntfs|
-        ntfs.blk_devices.each do |blk_device|
-          if Storage::partition?(blk_device)
-            if Storage::to_partition(blk_device).type == Storage::PartitionType_PRIMARY
-              return true if ntfs.detect_content_info.windows?
-            end
-          end
-        end
-      end
-      false
     end
 
     # While the rest of the installation dialogs have enough room
--- a/timezone/src/clients/inst_timezone.rb
+++ b/timezone/src/clients/inst_timezone.rb
@@ -20,17 +20,16 @@
 # ------------------------------------------------------------------------------
 # Client for initial timezone setting (part of installation sequence)
 # Author:	Jiri Suchomel <jsuchome@suse.cz>
-# $Id$
+
+require "storage"
+require "y2storage"
+
 module Yast
   class InstTimezoneClient < Client
     def main
       Yast.import "UI"
       Yast.import "GetInstArgs"
       Yast.import "Stage"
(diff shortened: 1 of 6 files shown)


mvidner:
It seems we have missed to fix [the check in timezone_proposal](https://github.com/aschnell/yast-country/blob/9fba6b143825e55b5eed71d9a0e61b26c65e7d80/timezone/src/clients/timezone_proposal.rb#L48).

aschnell:
Oh, that code was not DRY.

timezone/src/clients/inst_timezone.rb
```diff
@@ -20,39 +20,46 @@
...
-=end
 
       full_size_timezone_dialog
     end
 
+    # Checks whether the system has Windows installed on a primary partition with NTFS
+    def system_has_windows?
+      storage = Y2Storage::StorageManager.instance
+      Storage::Ntfs::all(storage.probed).each do |ntfs|
+        ntfs.blk_devices.each do |blk_device|
```
mvidner: Stupid question: `ntfs.blk_devices`? What is `ntfs` if it can have (belong to) multiple devices?

aschnell: In general filesystems can use several devices, e.g. btrfs or ext4 (separate journal device). For NTFS using only one of the BlkDevices would likely be fine but on the otherhand it does not make the code simpler.

aschnell:
I have adapted the code in timezone_proposal.rb.

It's unclear to me when that code is use so I don't know how to test it.

timezone/src/modules/Timezone.rb
```diff
@@ -984,6 +986,23 @@ def Summary
...
+    def system_has_windows?
+      storage = Y2Storage::StorageManager.instance
+      Storage::Ntfs::all(storage.probed).each do |ntfs|
+        ntfs.blk_devices.each do |blk_device|
+          if Storage::partition?(blk_device)
+            if Storage::to_partition(blk_device).type == Storage::PartitionType_PRIMARY
+              return true if ntfs.detect_content_info.windows?
+            end
+          end
+        end
```
ancorgs: I fully agree with the fact that libstorage-ng is programmed in a very agnostic way. A filesystem can spread on several block devices and all those kind of (not so) crazy combinations. But I'm not sure if I like the idea of having so much "low level" code in yast2-country. Maybe yast2-storage-ng should provide some more high level methods. Otherwise, the internal structure of libstorage-ng (with things like all those casts that look so unnatural in Ruby) will spread all over the YaST code.

ancorgs: I mean, in a random YaST module, checking if a devicegraph contains a windows installation (a kind of simple question) should not imply four occurrences of ``Storage::`.

aschnell: If you provide some examples of high-level functions we can discuss them.

But replace complexity by lack of clarity is not what we should aim for.

ancorgs: I believe that using that using `Y2Storage::Refinements::DevicegraphLists` the code can be rewritten as:
```ruby
def system_has_windows?
  devicegraph = Y2Storage::StorageManager.instance.probed
  prim_parts = devicegraph.partitions.with(type: Storage::PartitionType_PRIMARY)
  prim_parts.filesystems.with(type: Storage::FsType_NTFS).any? { |fs| fs.detect_content_info.windows? }
end
```

ancorgs:
@aschnell With travis is not enough to add a dependency to `BuildRequires`. You also has to add buildrequires explicitly in `.travis.yml`. Like this
```
...travis_setup.sh  -p "rake yast2-devtools yast2-testsuite yast2 yast2-ruby-bindings yast2-perl-bindings yast2-storage-ng" -g ...
```

timezone/src/modules/Timezone.rb
aschnell: OK, apparently I cannot write code in Ruby. I'll move the card back to the backlog.

mvidner: @aschnell if you do not give up, I will not give up on C++11! :-)
@ancorgs This is much clearer. In part because it removes the counterintuitive move from a fs to a block device, but also because it expresses queries and names the results. May I also suggest

```rb
ntfs_fss = prim_parts.filesystems.with(type: Storage::FsType_NTFS)
ntfs_fss.any? { |fs| fs.detect_content_info.windows? }
```

ancorgs: For some reasons (explained in the other pull request), I would like to merge this https://github.com/yast/yast-storage-ng/pull/92 and change the code to something like
```ruby
def system_has_windows?
  !disk_analyzer.windows_partitions.empty?
end

def disk_analyzer
  @disk_analyzer ||= begin
    devicegraph = Y2Storage::StorageManager.instance.probed
    Y2Storage::DiskAnalyzer(devicegraph)
  end
end
```

PR #90, opened 2016-09-27
Readded check for Windows based on storage-ng (DiskAnalyzer)

ancorgs:
Replaces #89 

It needs https://github.com/yast/yast-storage-ng/pull/92 to be integrated first.

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -24,10 +24,8 @@
 # Authors:	Klaus Kaempf <kkaempf@suse.de>
 #		Thomas Roelz <tom@suse.de>
 
-require "storage"
-require "y2storage"
-
 require "yast"
+require "y2storage"
 
 module Yast
   class TimezoneClass < Module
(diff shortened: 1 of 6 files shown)


timezone/src/modules/Timezone.rb
```diff
@@ -984,6 +984,12 @@ def Summary
       HTML.List(ret)
     end
 
+    # Checks whether the system has Windows installed
+    def system_has_windows?
+      win_partitions = disk_analyzer.windows_partitions.values.flatten
```
mvidner: windows_partitions returns `Hash{String => Array<::Storage::Partition>}`, so no `flatten` should be needed, right?

ancorgs: I believe is needed. `windows_partitions` could return
```ruby
{ "/dev/sda" => [] }
```
and
```ruby
{ "/dev/sda" => [] }.values.empty? #=> false
```

mvidner: Duh! Thanks for explaining the obvious!

PR #92, opened 2016-11-21
Bsc1010938

imobachgs:
* Specify tty devices when calling to `loadkeys` in the `Keyboard` module.
* Fixes https://bugzilla.suse.com/show_bug.cgi?id=1001245.

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -556,8 +556,9 @@ module Yast
         return false # Error
       end
 
-      # Console command...
-      @ckb_cmd = Ops.add("/bin/loadkeys ", @keymap)
+      # Console command. It specifies all tty devices (bsc#1010938)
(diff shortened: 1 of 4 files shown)


shundhammer:
LGTM

keyboard/src/modules/Keyboard.rb
```diff
@@ -556,8 +556,9 @@ def SetKeyboard(keyboard)
         return false # Error
       end
 
-      # Console command...
-      @ckb_cmd = Ops.add("/bin/loadkeys ", @keymap)
+      # Console command. It specifies all tty devices (bsc#1010938)
+      @tty_devices ||= Dir["/dev/tty*"].map { |d| "-C #{d}" }.join(" ")
```
imobachgs: @lslezak AFAIK, serial devices are attached to `/dev/ttyS*` devices, so we're covered here. On the other hand, I don't expect `/dev/tty*` devices to not exist.

lslezak: I see, then it's OK. Thanks!

PR #93, opened 2016-11-23
fixing default language

schubi2:


language/src/modules/Language.rb
```diff
@@ -802,7 +807,7 @@ def MakeProposal(force_reset, language_changed)
       Builtins.foreach(Pkg.GetAdditionalLocales) do |additional|
         # add the language for both kind of values ("cs" vs. "pt_PT")
         if !Builtins.contains(langs, additional)
-          additional = "en_US" if additional == "en"
+          additional = DEFAULT_FALLBACK_LANGUAGE if additional == "en"
```
imobachgs: We could consider using a hash for such mapping. But I guess is outside of the scope for this PBI.

package/yast2-country.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-country
-Version:        3.1.30
+Version:        3.1.30.1
```
imobachgs: The version is wrong. For SLE 12 SP2, you should use `3.1.31` (3-digits number) as `master` is currently `3.2.1` (it won't collide).

imobachgs:
LGTM

PR #96, opened 2016-12-01
add keyboard widget

jreidinger:
manually tested and works as expected.

### DIFF
--- a/keyboard/src/Makefile.am
+++ b/keyboard/src/Makefile.am
@@ -33,6 +33,7 @@ ylibdir = @ylibdir@/y2country
 ylib_DATA = \
   lib/y2country/widgets.rb
 
-EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA) $(ylib_DATA)
+EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) \
+  $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA) $(ylib_DATA)
 
 include $(top_srcdir)/Makefile.am.common
--- a/keyboard/src/Makefile.am
+++ b/keyboard/src/Makefile.am
@@ -29,6 +29,10 @@ ydata_DATA = \
 desktop_DATA = \
   desktop/keyboard.desktop
 
-EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA)
+ylibdir = @ylibdir@/y2country
+ylib_DATA = \
+  lib/y2country/widgets.rb
+
+EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA) $(ylib_DATA)
 
 include $(top_srcdir)/Makefile.am.common
--- a/keyboard/src/lib/y2country/widgets.rb
+++ b/keyboard/src/lib/y2country/widgets.rb
@@ -26,7 +26,7 @@ Yast.import "Keyboard"
 
 module Y2Country
   module Widgets
-    class KeyboardSelectionWidget < CWM::SelectionBox
+    class KeyboardSelection < CWM::SelectionBox
       # param default [String] ID for default keyboard layout if not selected.
       # Allowed values are defined in /usr/share/YaST2/data/keyboard_raw.ycp
       def initialize(default)
--- a/keyboard/src/lib/y2country/widgets.rb
+++ b/keyboard/src/lib/y2country/widgets.rb
@@ -25,50 +25,55 @@ require "cwm/widget"
 Yast.import "Keyboard"
 
 module Y2Country
-  class KeyboardSelectionWidget < CWM::SelectionBox
-    def initialize
-      textdomain "country"
-    end
+  module Widgets
+    class KeyboardSelectionWidget < CWM::SelectionBox
+      # param default [String] ID for default keyboard layout if not selected.
+      # Allowed values are defined in /usr/share/YaST2/data/keyboard_raw.ycp
+      def initialize(default)
+        textdomain "country"
+        @default = default
+      end
 
-    def label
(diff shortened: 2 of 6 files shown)


imobachgs:
In general, it looks good.

keyboard/src/lib/y2country/widgets.rb
```diff
@@ -0,0 +1,76 @@
...
+    def handle
+      Yast::Keyboard.Set(value)
+    end
+
+    def store
+      handle
+    end
+
+    def items
+      # a bit tricky as method return incompatible data
```
mvidner: OK, then let's add a yard comment to that method documenting the data. I'll do that.
Hmm, the method is used as an argument for a SelectionBox. Maybe we should fix [CWM::SelectionBox](http://www.rubydoc.info/github/yast/yast-yast2/CWM%2FSelectionBox) to accept that data format?

keyboard/src/lib/y2country/widgets.rb
```diff
@@ -0,0 +1,76 @@
...
+      res = Yast::Keyboard.GetKeyboardItems
+      res.map! do |item|
+        values = item.params
+        id, name, *_ = values
+        id = id.params.first
+        [id, name]
```
mvidner: Sorry to pick on implementation details, but since I had to understand the format differences anyway, here's an IMHO more readable version of the loop:
```rb
res.map do |item_term|  # non mutating map
  id_term, name, _enabled = item_term.params
  [id_term.params.first, name]
end

File: keyboard/src/lib/y2country/widgets.rb [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+# To contact Novell about this file by physical or electronic mail, you may find
+# current contact information at www.novell.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "cwm/widget"
+
+Yast.import "Keyboard"
+
+module Y2Country
```
mvidner: Sorry, I don't remember the naming scheme we have discussed in summer. Have you consulted it? (If yes, where can find it?)
In Users you have used [`module Users`](https://github.com/yast/yast-users/pull/127/files#diff-fa2036669a7ab51839a0ddf0e0b39295R32) which is inconsistent.

jreidinger: yes, because users already have lib dir with existing naming scheme, so I do not want to introduce two namespace in one module. But this is new module and according to new naming scheme ( still not yet officially annonced @imobachgs ?? ) this is name we should use. I explicitly ask before writting this.

keyboard/src/lib/y2country/widgets.rb
jreidinger: well, we can add ability to https://github.com/yast/yast-yast2/blob/5762181d62762816a73fc040362c1efb5d97deed/library/cwm/src/lib/cwm/widget.rb#L362 but it a  bit leak abstraction of widget. As goal of that object is to hide implementation details and ideally work with clean data which can allow us in future to change e.g. yui-bindings to nature libyui bindings.

keyboard/src/lib/y2country/widgets.rb
jreidinger: ok, but I will leave at least `id = id.params.first` to have there less magic and better understandable variable.

keyboard/src/Makefile.am
```diff
@@ -29,6 +29,10 @@ ydata_DATA = \
 desktop_DATA = \
   desktop/keyboard.desktop
 
-EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA)
+ylibdir = @ylibdir@/y2country
+ylib_DATA = \
+  lib/y2country/widgets.rb
+
+EXTRA_DIST = $(module_DATA) $(client_DATA) $(ynclude_DATA) $(scrconf_DATA) $(schemafiles_DATA) $(ydata_DATA) $(desktop_DATA) $(ylib_DATA)
```
mvidner: please wrap the line

keyboard/src/lib/y2country/widgets.rb
mvidner: OK, so the latest draft of the naming scheme is https://gist.github.com/imobachgs/d90940315bd0471ee00b2a68ba22fbe5 , the current version being https://gist.github.com/imobachgs/d90940315bd0471ee00b2a68ba22fbe5/b239d350a11a4500b510acc9c172f238e4df1626 .

According to the scheme, this class should be named `Y2Country::Widgets::KeyboardSelection`.

keyboard/src/lib/y2country/widgets.rb
```diff
@@ -0,0 +1,74 @@
...
+    def label
+      # widget label
+      _("&Keyboard Layout")
+    end
+
+    def init
+      if Yast::Keyboard.user_decision
+        self.value = Yast::Keyboard.current_kbd
+      else
+        self.value = "english-us"
```
mvidner: This seems out of place. The rest of the code does not care about how a keyboard ID looks like or what the default is. Can we inject it from the outside?

BTW the IDs are yast specific, defined in [/usr/share/YaST2/data/keyboard_raw.ycp](https://github.com/yast/yast-country/blob/a0bf4289fb27795da0840a4cbff724e82a19f8ac/keyboard/src/data/keyboard_raw.ycp)

keyboard/src/lib/y2country/widgets.rb
mvidner: No, it is not leaking implementation details of the CWM widget. It is sucking out implementation details out of libyui, for convenience :)
I don't insist.

keyboard/src/lib/y2country/widgets.rb
```diff
@@ -0,0 +1,79 @@
...
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "cwm/widget"
+
+Yast.import "Keyboard"
+
+module Y2Country
+  module Widgets
+    class KeyboardSelectionWidget < CWM::SelectionBox
```
mvidner: [The scheme](https://gist.github.com/imobachgs/d90940315bd0471ee00b2a68ba22fbe5#file-proposal-md) is not actually clear on this. but as I understood it, this class should be named `Y2Country::Widgets::KeyboardSelection`, not `Y2Country::Widgets::KeyboardSelectionWidget`.
@imobachgs we should make that clear.

jreidinger:
thanks

PR #97, opened 2016-12-02
Set timezone as read-only

imobachgs:
Implements [Force timezone "UTC"](https://trello.com/c/1Mrvm2EH/781-5-casp-force-timezone-utc) PBI. It's not finished yet but feedback is welcome.

As a first approach, I've decided to add a new option to the control file readonly_timezone (perhaps I need a better name). If it's set to true, the timezone will be initialized to UTC and the Timezone.Set method will refuse to change it. I consider it a barrier to avoid accidental changes.

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -29,6 +29,8 @@ require "yast"
 
 module Yast
   class TimezoneClass < Module
+    include Yast::Logger
+
     def main
       textdomain "country"
 
@@ -163,6 +165,10 @@ module Yast
 
       # remember if /sbin/hwclock --hctosys was called, it can be done only once (bnc#584484)
       @systz_called = false
+
+      # timezone is read-only
+      @readonly_timezone = nil
+
       Timezone()
     end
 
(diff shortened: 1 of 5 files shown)


timezone/src/modules/Timezone.rb
```diff
@@ -431,9 +443,14 @@ def Timezone
       @hwclock = "-u"
       if Stage.initial && !Mode.live_installation
         # language --> timezone database, e.g. "de_DE" : "Europe/Berlin"
-        lang2tz = get_lang2tz
+        new_timezone =
+          if readonly
+            "UTC"
```
imobachgs: The plan is to honor the `<timezone/>` element in the control file.

timezone/src/modules/Timezone.rb
```diff
@@ -245,6 +251,12 @@ def get_zonemap
     # @return	the number of the region that contains the timezone
     #
     def Set(zone, really)
+      # Do not update the timezone if it's forced and it was already set
+      if readonly && !@timezone.empty?
+        log.info "Timezone is read-only and cannot be changed"
+        return -1 # Not sure about this
```
imobachgs: I'm returning a `-1` to avoid changing the method's signature as some code could rely on this.

schubi2:
LGTM

schubi2:
LGTM again

imobachgs:
Thanks!

PR #99, opened 2016-12-12
Set timezone as read-only

imobachgs:
* Merge #97 into CASP branch.

### DIFF
--- a/timezone/src/clients/inst_timezone.rb
+++ b/timezone/src/clients/inst_timezone.rb
@@ -23,6 +23,8 @@
 # $Id$
 module Yast
   class InstTimezoneClient < Client
+    include Yast::Logger
+
     def main
       Yast.import "UI"
       Yast.import "GetInstArgs"
@@ -34,6 +36,12 @@ module Yast
 
       @args = GetInstArgs.argmap
       @args["first_run"] = "yes" unless @args["first_run"] == "no"
+
+      if Timezone.readonly
+        # Do not run if timezone is readonly
+        log.info "Timezone is read-only for this product so the inst_timezone client is skipped"
+        return GetInstArgs.going_back ? :back : :next
+      end
 
       if Stage.initial &&
           Ops.greater_than(
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -29,6 +29,8 @@ require "yast"
 
 module Yast
   class TimezoneClass < Module
+    include Yast::Logger
+
     def main
       textdomain "country"
 
@@ -163,6 +165,10 @@ module Yast
 
       # remember if /sbin/hwclock --hctosys was called, it can be done only once (bnc#584484)
       @systz_called = false
+
+      # timezone is read-only
+      @readonly = nil
+
       Timezone()
     end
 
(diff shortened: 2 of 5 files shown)


timezone/src/clients/inst_timezone.rb
```diff
@@ -35,6 +37,12 @@ def main
       @args = GetInstArgs.argmap
       @args["first_run"] = "yes" unless @args["first_run"] == "no"
 
+      if Timezone.readonly
+        # Do not run if timezone is readonly
+        log.info "Timezone is read-only for this product so the inst_timezone client is skipped"
+        return GetInstArgs.going_back ? :back : :next
```
jreidinger: for this purpose we use `:auto` which do this `:back/next` automatic

timezone/src/modules/Timezone.rb
```diff
@@ -984,6 +1000,26 @@ def Summary
       HTML.List(ret)
     end
 
+    # Determines whether timezone is read-only for the current product
+    #
+    # @return [Boolean] true if it's read-only; false otherwise.
+    def readonly
+      @readonly ||= ProductFeatures.GetBooleanFeature("globals", "readonly_timezone")
```
jreidinger: using `||=` for boolean parameters is really bad idea.

timezone/src/modules/Timezone.rb
```diff
@@ -984,6 +1000,26 @@ def Summary
...
+    end
+
+    # Product's default timezone when it's not defined in the control file.
```
jreidinger: please mention FATE which define this default.

timezone/src/modules/Timezone.rb
```diff
@@ -245,20 +251,25 @@ def get_zonemap
     # @return	the number of the region that contains the timezone
     #
     def Set(zone, really)
-      zmap = get_zonemap
+      # Do not update the timezone if it's forced and it was already set
+      if (Mode.installation || Mode.update) && readonly && !@timezone.empty?
```
imobachgs: Yes, it tries to handle also the AutoYaST case. To be honest, initially I used `Stage.initial` but fortunately @schubi2 noticed that it could break AutoYaST.

timezone/src/clients/inst_timezone.rb
imobachgs: Oh, you're right. Thanks.

timezone/src/modules/Timezone.rb
```diff
@@ -984,6 +1000,26 @@ def Summary
...
+    def readonly
+      @readonly ||= ProductFeatures.GetBooleanFeature("globals", "readonly_timezone")
+    end
+
+    # Product's default timezone when it's not defined in the control file.
+    FALLBACK_PRODUCT_DEFAULT_TIMEZONE = "UTC"
+
+    # Determines the default timezone for the current product
+    #
+    # If not timezone is set, "UTC" will be used.
```
imobachgs: I introduced the constant after writing the documentation and I forgot to change it.

timezone/src/modules/Timezone.rb
jreidinger: yes, harmless. just caching do not work, but I think call is not so expensive.

timezone/src/modules/Timezone.rb
jreidinger: and what about firstboot scenario?

imobachgs: I think that we should allow to change the timezone in firstboot. The PBI refers to inst-sys and the installed system, but the user is allowed to change the timezone of a running system. I know that we could consider firstboot as part of the installation, but IMHO there's no point in asking for timezone during firstboot if you don't want the user to change it.

jreidinger:
LGTM

imobachgs:
Thanks!

PR #100, opened 2016-12-13
Improve readonly timezone

imobachgs:
Small improvements suggested during review of #99.


jreidinger:
LGTM, but why it is not only in CASP branch and already in SP2? I though we merge it later to SP2.

imobachgs:
For no particular reason. I always use the same flow: SP2->CASP->master. Maybe it makes sense when fixing bugs but not when implementing new features.

imobachgs:
Thanks, BTW :)

PR #104, opened 2016-12-19
Keyboard and console fonts per product

imobachgs:
Just a PoC not even tested.


console/src/modules/Console.rb
```diff
@@ -240,6 +241,26 @@ def Console
...
+    # @return [Hash] Console fonts map
+    def consolefonts
+      return @consolefonts if @consolefonts
+
+      # Read the file for this product
+      product_filename = "consolefonts_#{OSRelease.id}.ycp"
+      @consolefonts = WFM.Read(path(".local.yast2"), product_filename)
+
+      # Fallback
+      @consolefonts ||= WFM.Read(path(".local.yast2"), "consolefonts.ycp")
```
jreidinger: well, SLE12 do not have ID? why it depends on fallback?

keyboard/src/modules/Keyboard.rb
```diff
@@ -1522,6 +1518,23 @@ def enable_autorepeat
...
+  #
+  # The map can be read from two different files:
+  #
+  # * `keyboard_raw_PRODUCT.ycp` where PRODUCT is the downcased product's
+  #   short name. For example, `keyboard_raw_sles.ycp`.
+  # * `keyboard_raw.ycp` as a fallback.
+  #
+  # @return [Hash] Keyboards map
+  def all_keyboards
+    content = SCR.Read(path(".target.yast2"), "keyboard_raw_#{OSRelease.id}.ycp")
```
mchf: 1) I miss explicit OSRelease import in the file
2) I've tried OSRelease.id in irb and it stated "undefined method `id'"

jreidinger: 2. it is new call and imo add properly dependency, so it should be fine.

mchf: i see, thx

console/src/modules/Console.rb
```diff
@@ -240,6 +241,26 @@ def Console
       nil
     end
 
+    # Console fonts map
+    #
+    # The map can be read from two different files:
+    #
+    # * `consolefonts_PRODUCT.ycp` where PRODUCT is the downcased product's
+    #   short name. For example, `consolefonts_sles.ycp`.
+    # * `consolefonts.ycp` as a fallback.
```
jreidinger: +1, I am just not sure if such change is good idea for SP2

imobachgs: I would prefer to keep the .ycp version for SLE 12 SP2 and CASP. But, what about using YAML in `master`? Does it sound reasonable?

keyboard/src/modules/Keyboard.rb
imobachgs: 1. you're right. I'll add an explicit import.

console/src/modules/Console.rb
imobachgs: Just to minimize the amount of `consolefonts.ycp` files, so `casp` and `sles` will share the same file. On the other hand, if some release miss the `id` field, we'll have a fallback.

console/src/modules/Console.rb
```diff
@@ -246,6 +248,37 @@ def Console
...
+    # The map can be read from two different files:
+    #
+    # * `consolefonts_ID.ycp` where ID is the distribution identifier (as
+    #   specified in /etc/os-release). For example, `consolefonts_opensuse.ycp`.
+    # * `consolefonts.ycp` as a fallback.
+    #
+    # Associates languages with the following set of properties: font, unicode map,
+    # screen map and magic initialization.
+    #
+    #   consolefonts #=>
```
jreidinger: what about using `@example` tag? ouput will look much better.

imobachgs: Oh, I forgot the tag (I used it in the keyboard module example).

package/yast2-country.spec
```diff
@@ -28,7 +28,7 @@ License:        GPL-2.0
 #policy files for YaPI dbus interface
 Source1:        org.opensuse.yast.modules.yapi.time.policy
 Source2:        org.opensuse.yast.modules.yapi.language.policy
-BuildRequires:	perl-XML-Writer update-desktop-files yast2-testsuite yast2 yast2-perl-bindings
+BuildRequires:	perl-XML-Writer update-desktop-files yast2-testsuite yast2-perl-bindings
```
jreidinger: I suggest to run spec formatter on this spec as it looks a bit ugly. script is at /usr/lib/obs/service/format_spec_file

console/src/modules/Console.rb
mchf: fine with me

console/src/data/consolefonts.ycp
```diff
@@ -1,7 +1,20 @@
...
+ * Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ * ------------------------------------------------------------------------------
+ */
+
 /*
  * Maintainer:  Thomas Fehr <fehr@suse.de>
```
mchf: NP for master version ... who is thomas fehr? AFAIK he is not in SUSE.

imobachgs: @mchf We'll lose that information when we convert it to YAML :)

imobachgs:
@jreidinger @mchf Thanks both!

PR #107, opened 2016-12-22
Keyboard and console fonts per product

imobachgs:
Merge #105. I was working in migration from ycp to yaml, but I had some problems with Travis so I've decided to do it later.


imobachgs:
Thanks! I think it's ok. I've compared `consolefonts.ycp` with the one we ship in SLE 12 SP2 and `consolefonts_opensuse.ycp` with the Tumbleweed version and they match.

PR #112, opened 2017-01-30
Do not enforce basic keyboard X11Variant

ancorgs:
See https://bugzilla.suse.com/show_bug.cgi?id=1020082#c8 for the rationale.


keyboard/src/modules/Keyboard.rb
```diff
@@ -1551,6 +1539,23 @@ def all_keyboards
       # eval is necessary for translating the texts needed to be translated
       content ? Builtins.eval(content) : {}
     end
+
+    def call_set_x11_keymap
+      args = [@XkbLayout, @XkbModel, @XkbVariant, @XkbOptions]
+      return if args.all?(&:empty?)
+      
+      # According to localectl syntax, empty options must be entered as "" if
+      # there are more non-empty options afterwards and skipped otherwise
```
mchf: I had to read this several times ... and still didn't get it until I read the code ;-)

keyboard/src/modules/Keyboard.rb
```diff
@@ -1551,6 +1539,24 @@ def all_keyboards
       # eval is necessary for translating the texts needed to be translated
       content ? Builtins.eval(content) : {}
     end
+
+    def call_set_x11_keymap
+      args = [@XkbLayout, @XkbModel, @XkbVariant, @XkbOptions]
+      return if args.all?(&:empty?)
```
mchf: i guess that none of ```@Xkb....``` can be nil.

ancorgs: No, it cannot. They are all initialized to a string (to `""` in most cases).

mchf: i'll answer myself ... no it cannot

PR #120, opened 2017-03-23
Add console tests to Makefile

ancorgs:
The purpose of this pull request if to prove that `console_test.rb` fails in some situations. Maybe related to some recent bug reports.

### DIFF
--- a/language/test/Language_test.rb
+++ b/language/test/Language_test.rb
@@ -162,15 +162,33 @@ describe "Language" do
   end
 
   describe "#GetLocaleString" do
-    context "when the language is defined" do
-      it "returns the full language identifier" do
+    context "when using UTF-8" do
+      it "returns the full locale" do
         expect(subject.GetLocaleString("de_ZU")).to eq("de_ZU.UTF-8")
+      end
+
+      context "and the language is not found in the database" do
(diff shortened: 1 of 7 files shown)


language/test/Language_test.rb
```diff
@@ -160,4 +161,17 @@
...
+        expect(subject.GetLocaleString("de_ZU")).to eq("de_ZU.UTF-8")
+      end
+    end
+
+    context "when the language is not defined" do
+      it "returns the full language identifier" do
+        expect(subject.GetLocaleString("ma_MA")).to eq("ma_MA.UTF-8")
+      end
+    end
+  end
```
ancorgs: In these tests it not clear enough (to me, at first sight) where the `UTF-8` part comes from or if there is a possibility to add a different suffix or no suffix at all. Just looking at this, it looks like a method to append `.UTF-8` to whatever string it receives as parameter.

imobachgs: Because it's your own PR :) I'll extend the text to use a different `@use_utf8` value and it will make it more clear.

ancorgs:
The tests are better now... it's still confusing that the method does the same no matter if the language is in the database... but that's life with this old code. LGTM

PR #122, opened 2017-05-17
remove last remaining old storage reference

wfeldt:
'timezone' now also runs on the installed system. (It will crash when
run as non-root user due to storage-ng requiring root.)

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -1075,10 +1075,7 @@ module Yast
   protected
 
     def disk_analyzer
-      @disk_analyzer ||= begin
-        devicegraph = Y2Storage::StorageManager.instance.y2storage_probed
-        Y2Storage::DiskAnalyzer.new(devicegraph)
-      end
+      Y2Storage::StorageManager.instance.probed_disk_analyzer
     end
   end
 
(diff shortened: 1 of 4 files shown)


ancorgs:
LGTM.

Not 100% related, but since we are here, it would be cool if we use the opportunity to change this in the `Timezone` module.
From:
```ruby
def disk_analyzer
  @disk_analyzer ||= begin
    devicegraph = Y2Storage::StorageManager.instance.y2storage_probed
    Y2Storage::DiskAnalyzer.new(devicegraph)
  end
end
```
To:
```ruby
def disk_analyzer
  Y2Storage::StorageManager.instance.probed_disk_analyzer
end
```
On one hand, it will improve caching of results during installation. On the other, it will save us from revisiting yast-country again when we remove `StorageManager#y2storage_probed` (that is a temporary method).

timezone/src/modules/Timezone.rb
```diff
@@ -1075,10 +1075,7 @@ def product_default_timezone
   protected
 
     def disk_analyzer
-      @disk_analyzer ||= begin
-        devicegraph = Y2Storage::StorageManager.instance.y2storage_probed
-        Y2Storage::DiskAnalyzer.new(devicegraph)
-      end
+      devicegraph = Y2Storage::StorageManager.instance.probed_disk_analyzer
```
ancorgs: The `devicegraph =` part should not be there.

wfeldt: ooops

PR #125, opened 2017-06-19
bsc#1023774: YaST shouldn't write directly into vconsole.conf

gilsonsouza:


keyboard/src/modules/Keyboard.rb
```diff
@@ -885,9 +885,11 @@ def Save
       )
       SCR.Write(path(".sysconfig.keyboard"), nil) # flush
 
-      # Write systemd settings for console
-      SCR.Write(path(".etc.vconsole_conf.KEYMAP"), @keymap.gsub(/(.*)\.map\.gz/, '\1'))
-      SCR.Write(path(".etc.vconsole_conf"), nil) # flush
+      cmd = "/usr/bin/localectl set-keymap #{@keymap.gsub(/(.*)\.map\.gz/, '\1')}"
```
ancorgs: Since we set the x11 keymap and the console one separately (each one with its own call to `localectl`), I would say `--no-convert` is missing here.

gilsonsouza: I see! I'll change and test it! Thx!

PR #127, opened 2017-07-27
Content removal

jreidinger:


language/src/modules/Language.rb
```diff
@@ -1318,39 +1318,10 @@ def GetLanguageItems(kind)
...
-              "Enable online repositories later in order to get the appropriate support\n" +
-              "for this language.\n"
-          )
-        )
-      end
-
-      nil
+    # @deprecated do nothing
+    def CheckLanguagesSupport(_selected_language)
+      log.warn "Called check for language support, but it do nothing"
```
lslezak: It **does** nothing.

package/yast2-country.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul 27 11:19:13 UTC 2017 - jreidinger@suse.com
+
+- No longer read /content file from media (FATE#322386)
```
lslezak: Please mention from which place it is read now.

PR #128, opened 2017-08-03
Casp merge sp2

jreidinger:


package/yast2-country.spec
```diff
@@ -39,13 +39,13 @@ BuildRequires:  yast2-core >= 3.1.12
 # RSpec extensions for YaST
 BuildRequires:  yast2-ruby-bindings >= 3.1.26
 # OSRelease.id
-BuildRequires:  yast2 >= 3.1.211
+BuildRequires:  yast2 >= 3.1.210.2
```
teclator: Although it doesn't hurt It should continue requiring 3.1.211 version instead of 3.1.210.2. Isn't it?

jreidinger: well, quite tricky question. If we require 211, then if we release update for this package for SP2, it will work fine there, but for caasp if we do not increase also version for it, then for caasp we did not have yast2 in proper version.

teclator: Yep, that is true

PR #131, opened 2017-08-10
Merge sp3 cherrypick

ancorgs:
We cherry-picked a .gitignore change from master to the newly created SLE-12-SP3 branch. So this is the corresponding merge to sync Git.


ancorgs:
As pointed by @jreidinger, a merge from SP3 to master is expected very soon (right after merging SP2 to SP3), so this is not that important.

PR #134, opened 2017-08-14
Merge branch 'storage-ng' into 'master' and release 3.3.1

wfeldt:


timezone/src/clients/timezone.rb
```diff
@@ -137,22 +137,9 @@ def TimezoneSequence
...
-        partitions = Convert.convert(
-          WFM.call("wrapper_storage", ["GetWinPrimPartitions", [targets]]),
-          :from => "any",
-          :to   => "list <map>"
-        )
-        if partitions != nil && Ops.greater_than(Builtins.size(partitions), 0)
-          Timezone.windows_partition = true
-          Builtins.y2milestone("windows partition found")
-        end
+      if Timezone.system_has_windows?
```
shundhammer: Just a matter of style and technically not really a part of this merge effort, but shouldn't this system_has_windows? method rather do its own caching and only really probe if the probing wasn't done before and other wise return the cached value? Explicitly setting that cached value here feels wrong to me.

This shouldn't keep us from merging, but let's think about that.

PR #139, opened 2017-09-06
Language selection for caasp

jreidinger:


keyboard/src/lib/y2country/widgets.rb
```diff
@@ -23,13 +23,15 @@
...
 Yast.import "Keyboard"
+Yast.import "Language"
 
 module Y2Country
   module Widgets
     # Common parts for {KeyboardSelection} and {KeyboardSelectionCombo}.
     module KeyboardSelectionBase
-      # param default [String] ID for default keyboard layout if not selected.
-      # Allowed values are defined in /usr/share/YaST2/data/keyboard_raw.ycp
+      # param default [String] ID for default keyboard layout if not selected
```
imobachgs: `@param` instead of `param`.

language/src/lib/y2country/widgets/language_selection.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      # Handle value changes
+      def handle
+        return if value.nil? || value == default
+        Yast::Timezone.ResetZonemap
+        Yast::Language.Set(value)
+        Yast::Language.languages = Yast::Language.RemoveSuffix(value)
+        @default = value
+        return nil unless @emit_event
+        return nil if Yast::Mode.config
```
imobachgs: np: Why not merge both ifs? If the current implementation looks more readable to you, just keep it as it is.

language/src/lib/y2country/widgets/language_selection.rb
```diff
@@ -0,0 +1,128 @@
...
+
+        switch_language
+        :redraw
+      end
+
+      # Store widget value
+      def store
+        handle
+
+        switch_language if !@emit_event && !Yast::Mode.config
```
imobachgs: Why not `unless @emit_event || Yast::Mode.Config`? BTW, see my comment above about merging conditions.

jreidinger: well, I really dislike unless with more complex condition. because `unless @emit_event || Yast::Mode.Config` is in my head `if !(@emit_event || Yast::Mode.Config)` which I need to translate to `if !@emit_event && !Yast::Mode.Config` so it is three steps computation, so I prefer to be explicit and see it on first sight.
Maybe behavior from languages which does not have unless. So `unless @emit_event` is fine for me as it is simple `if !@emit_event` and is trivial to see, but for more complex expression I have to do computation in my head and it decrease readiness in my head.

imobachgs: If you find it more readable, then keep it as it is.

test/test_helper.rb
```diff
@@ -14,7 +14,7 @@
   end
 
   # for correct coverage report we need to load all ruby files
-  Dir["#{root_location}/*/src/{module,include,lib}/**/*.rb"].each { |f| require_relative f }
+  SimpleCov.track_files("#{root_locaiton}/**/src/**/*.rb")
```
imobachgs: `root_locaiton` looks like a typo.

jreidinger:
coverage decreased because I fixed way how it counting rb files. Now it includes also clients, which is quite huge in this module.

PR #143, opened 2017-09-20
Remove dependency on yast-storage (-ng)

shundhammer:
https://trello.com/c/JLyn7xAV/721-1-yast-country-make-dependency-on-storage-ng-optional

Now soft-failing if y2storage is not available: Just write a warning to the log and assume the hardware clock to be running on UTC, i.e. just the same as when there is no windows partition.

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -25,7 +25,12 @@
 #		Thomas Roelz <tom@suse.de>
 
 require "yast"
-require "y2storage"
+
+begin
+  require "y2storage"
+rescue LoadError
+  # Ignore y2storage not being available (bsc#1058869)
+end
 
 module Yast
   class TimezoneClass < Module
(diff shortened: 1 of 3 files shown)


timezone/src/modules/Timezone.rb
```diff
@@ -1002,8 +1007,16 @@ def Summary
...
+      begin
+        win_partitions = disk_analyzer.windows_partitions
+        !win_partitions.empty?
+      rescue NameError => ex
+        # bsc#1058869: Don't enforce y2storage being available
+        log.warn("Caught #{ex}")
+        log.warn("No storage-ng support - not checking for a windows partition")
+        log.warn("Assuming UTC for the hardware clock")
+        false # No windows partition found
+      end
```
jreidinger: btw this can be written as

```
def system_has_windows?
  require "y2storage"
  disk_analyzer = Y2Storage::StorageManager.instance.probed_disk_analyzer
  win_partitions = disk_analyzer.windows_partitions
  !win_partitions.empty?
rescue LoadError
  log.warn("No storage-ng support - not checking for a windows partition")
  false
end
```

which quite simplify all these stuff ( disk analyzer method can be removed and similar that require on top )

PR #144, opened 2017-10-09
Keyboard rewrite: Load keyboard layouts

jyeray:
This is for https://github.com/openSUSE/mentoring/issues/79, "Re-implement the YaST keyboard management in a proper object oriented way"

This first step is only to load the layouts on the system, add it a "description" and show in a simple user interface.


keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
```diff
@@ -0,0 +1,26 @@
...
+      "fr" => "French",
+      "us" => "English (US)"
+    }
+
+    def initialize(code, description)
+      @code = code
+      @description = description
+    end
+
+    def self.load
```
imobachgs: I would prefer a name like `all` so the API will look like like:

```ruby
Y2Keyboard::KeyboardLayout.all # I want all the layouts
```

After all, I do not want to know if you want to load them, or if you have a hardcoded list, or whatever...

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
```diff
@@ -0,0 +1,26 @@
...
+      raw_layouts = Cheetah.run("localectl", "list-keymaps", stdout: :capture)
```
imobachgs: What about using [memoization](https://en.wikipedia.org/wiki/Memoization) here? You can just cache the value and do not call to `localectl` again.

```ruby
def all
  return @all if @all
  # do things...
  @all = code_with_description.map { ... }
end
```

After all, we do no expect the list of locales to change so often.

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
```diff
@@ -0,0 +1,26 @@
...
+      layout_codes = raw_layouts.lines.map { |string| string.split.first }
```
imobachgs: I would use a shorter name (`codes` should be enough). About the `string` argument for the block, the name tell me nothing. You could use `code` (I tend to use a single letter).

On the other hand, if you only want to remove the "\n", you should better use `strip`.

```ruby
codes = raw_layouts.lines.map { |c| c.strip }
```

And you can turn this into:

```ruby
codes = raw_layouts.map(&:strip)
```

keyboard_rewrite/test/keyboard_layout_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it "initialize the layout description" do
+      layout_list = ["es"]
+      given_layouts(layout_list)
+
+      expect(load_keyboard_layouts.first.description).to eq("Spanish")
+    end
+
+    it "does not returns layouts that not have description" do
```
imobachgs: "does not *return* layouts that *does* not have description"

keyboard_rewrite/test/test_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+  if ENV["TRAVIS"]
+    require "coveralls"
+    SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
+      SimpleCov::Formatter::HTMLFormatter,
+      Coveralls::SimpleCov::Formatter
+    ]
+  end
+end
+
+def given_layouts(layouts_to_return)
```
imobachgs: I would suggest to move these helper methods to its own module and including them using the RSpec config method. See [this example](https://github.com/yast/yast-installation/blob/5ae24ff36f370f41b84bed5ae5f3107bc0552f87/test/test_helper.rb#L59).

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
jyeray: I think that in this context it is not necessary, because it would be used only once in the module lifetime, and on other hand that will add more complexity to the tests.

jyeray:
Thanks @imobachgs I will take a look to the existing widgets in keyboard and language to include it another pull request.

keyboard_rewrite/src/lib/y2_keyboard/clients/main.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "../dialogs/layout_selector"
+
+module Y2Keyboard
+  module Clients
+    class Main
```
imobachgs: np: Why `Main`? After all, when you are invoking the client, you are using `yast2 keyboard`.

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
```diff
@@ -18,8 +18,8 @@ def initialize(code, description)
 
     def self.all
       raw_layouts = Cheetah.run("localectl", "list-keymaps", stdout: :capture)
-      layout_codes = raw_layouts.lines.map { |string| string.split.first }
-      codes_with_description = layout_codes.select { |code| LAYOUT_CODE_DESCRIPTIONS.key?(code) }
+      codes = raw_layouts.lines.map(&:strip)
+      codes_with_description = codes.select { |code| LAYOUT_CODE_DESCRIPTIONS.key?(code) }
```
imobachgs: np: I prefer to use single letter variables when using one-line blocks. See [our Ruby style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md#parameters).

imobachgs:
I've launched Travis tests again and now they are failing. The main problem you might face is that yast2-country has evolved quite a bit since this PR was open (version 4.0.20 against 4.0.2). It would be a good idea rebasing (or merging `master` back, as you prefer).

keyboard_rewrite/src/lib/y2_keyboard/clients/main.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "../dialogs/layout_selector"
```
ancorgs: I assume the usage of `require_relative` is just temporary until we have everything set up.

yast2-country is one of the oldest YaST modules. It still uses Makefile and it sucks regarding location of files, so `require_relative` is acceptable for the time being. But we should kill it at some point.

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
```diff
@@ -0,0 +1,26 @@
...
+      "us" => "English (US)"
+    }
+
+    def initialize(code, description)
+      @code = code
+      @description = description
+    end
+
+    def self.all
+      raw_layouts = Cheetah.run("localectl", "list-keymaps", stdout: :capture)
```
ancorgs: I like the initial API and implementation of this class... to be the initial one. :wink:

Take into account that in a close future we have to make this relatively independent of systemd (likely using some pattern like Strategy). But doing it now would be premature, of course.

keyboard_rewrite/test/helpers.rb
```diff
@@ -0,0 +1,12 @@
+module Helpers
```
ancorgs: Very dangerous name for a module. I can hear the name collisions coming. :wink:

keyboard_rewrite/src/lib/y2_keyboard/clients/main.rb
jyeray: When you suggest me [this change](https://github.com/yast/yast-country/pull/144/files/0599b6009b6546ed7c0c78f20a27368778fbd2a2#diff-20cc8e98fac2471531f1ee0b6fcd12dd), I take [this]( https://github.com/yast/yast-storage-ng/blob/master/src/clients/partitioner.rb) as example where the client is also named ```Main``` 
There is no other reason, so I will rename it.

keyboard_rewrite/src/lib/y2_keyboard/keyboard_layout.rb
jyeray: Yes, I have that in mind, but the API is to small yet, so I prefer to keep it simple for now.

keyboard_rewrite/src/lib/y2_keyboard/clients/main.rb
jyeray: Yeah it would be ideal to use require instead of require_relative, I will write it in a technical debt file, to remember it.

jyeray:
I have review all comments, only one thing is pending, the coverage results. 

I have taken a look to travis log and it seems to execute all test, the existing ones, and the new ones. But as result of coverage shows [this](https://travis-ci.org/yast/yast-country/jobs/361329833#L1354). 

In coveralls only files added in this branch appear as "tested", I dont know why :confused:

jyeray:
Fixed the problems with the coverage :smiley:

PR #145, opened 2017-10-18
[WIP] adapt to use systemd-firstboot in installation for keyboard (bsc#1046…

jreidinger:
https://bugzilla.opensuse.org/show_bug.cgi?id=1046436

- [x] changes
- [x] manual testing

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -886,7 +886,7 @@ module Yast
       )
       SCR.Write(path(".sysconfig.keyboard"), nil) # flush
 
-      keymap = @keymap.gsub(/(.*)\.map\.gz/, '\1')
+      keymap = @keymap.chomp(".map.gz")
       cmd = if Stage.initial
         "/usr/bin/systemd-firstboot --root #{Installation.destdir} --keymap '#{keymap}'"
       else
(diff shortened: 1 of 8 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -885,10 +885,18 @@ def Save
       )
       SCR.Write(path(".sysconfig.keyboard"), nil) # flush
 
-      cmd = "/usr/bin/localectl --no-convert set-keymap #{@keymap.gsub(/(.*)\.map\.gz/, '\1')}"
+      cmd = if Stage.initial
+        # lets hardcode /mnt there as we do not want to add dependency on yast2-installation
+        # and its Yast::Installation.destdir
+        "/usr/bin/systemd-firstboot --root /mnt --keymap '#{@keymap.gsub(/(.*)\.map\.gz/, '\1')}'"
```
lslezak: NP: I know, it was already there, but that `gsub` looks too complicated to me. What about simpler `@keymapgsub(/\.map\.gz/,"")` ?

test/test_helper.rb
```diff
@@ -2,6 +2,9 @@
 inc_dirs = Dir.glob("#{root_location}/*/src")
 ENV["Y2DIR"] = inc_dirs.join(":")
 
+ENV["LANG"] = "C"
```
lslezak: I'd prefer `en_US.UTF-8` I think there could be possibly some problems with UTF-8 strings. Moreover using non-UTF8  is probably not supported and should not be normally used in the target systems.

mvidner: Right, lets use that.

lslezak:
Um, the fix did not fully work for me in the manual test I did, but at least the log shows the correct command was called. So lets submit this and test it together with the systemd pieces...

PR #147, opened 2017-10-25
Display an error popup when saving the keyboard config fails

lslezak:
- Related to [bsc#1046436](https://bugzilla.suse.com/show_bug.cgi?id=1046436)
- 4.0.5


lslezak:
Um, for some reason Travis fails downloading the `yastdevel/ruby` Docker image... :-(

But my local test run was OK.

PR #148, opened 2017-10-31
Fixed saving X11 keyboard (bsc#1065264)

lslezak:
- The `localectl` tool can create the X11 configuration from the console setting (when the `--no-convert` option is not used), no need to reimplement the functionality in YaST
- See the details in the [man page](https://www.freedesktop.org/software/systemd/man/localectl.html#set-keymap%20MAP%20%5BTOGGLEMAP%5D)
- 4.0.6

### DIFF
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 31 14:41:28 UTC 2017 - lslezak@suse.cz
+
+- Let the "localectl" to tool create the X11 keyboard configuration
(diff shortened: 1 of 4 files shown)


package/yast2-country.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 31 14:41:28 UTC 2017 - lslezak@suse.cz
+
+- Let the "localectl" to tool create the X11 keyboard configuration
```
jreidinger: sound strange "to tool create" so maybe "tool to create"

PR #149, opened 2017-11-07
Update for setting language via localectl

jsrain:
https://fate.suse.com/319454

### DIFF
--- a/language/src/lib/y2country/language_dbus.rb
+++ b/language/src/lib/y2country/language_dbus.rb
@@ -21,7 +21,7 @@
 
 # File:	modules/Language.ycp
 # Module:	Language
-# Summary:	This module does all language related stuff:
+# Summary:	DBus interface for localed.conf
 # Authors:	Klaus Kaempf <kkaempf@suse.de>
 #		Thomas Roelz <tom@suse.de>
 # Maintainer:  Jiri Suchomel <jsuchome@suse.cz>
@@ -32,10 +32,14 @@ require "yast"
 module Y2Country
     extend Yast::Logger
 
+    # Read via DBus the locale settings
+    # @return hash of locale variables
     def read_locale_conf
       begin
         require "dbus"
       rescue
+        # inst-sys (because of constructor)
+        log.info("DBus module not available")
         return nil
       end
       localed_conf = {}
--- a/language/src/lib/y2country/language_dbus.rb
+++ b/language/src/lib/y2country/language_dbus.rb
@@ -45,6 +45,7 @@ module Y2Country
       sysbus = DBus.system_bus
       locale_service   = sysbus["org.freedesktop.locale1"]
       locale_object    = locale_service.object "/org/freedesktop/locale1"
+      # following line not necessary with ruby-dbus >= 0.13.0
       locale_object.introspect # needed, https://github.com/mvidner/ruby-dbus/issues/28
       locale_interface = locale_object["org.freedesktop.locale1"]
       locales          = locale_interface["Locale"]
--- a/language/src/lib/y2country/language_dbus.rb
+++ b/language/src/lib/y2country/language_dbus.rb
@@ -28,12 +28,16 @@
 #
 # $Id$
 require "yast"
-require "dbus"
 
 module Y2Country
     extend Yast::Logger
 
     def read_locale_conf
+      begin
+        require "dbus"
+      rescue
+        return nil
+      end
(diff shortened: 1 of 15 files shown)


language/src/modules/Language.rb
```diff
@@ -545,6 +547,22 @@ def SetDefault
       nil
     end
 
+    # read the localed.conf language
+    def ReadLocaleConfLanguage
+      return nil if Mode.testsuite
+      @localed_conf  = Y2Country.read_locale_conf
+      return nil if @localed_conf.nil?
+      local_lang = @localed_conf["LANG"]
+      pos = Builtins.findfirstof(local_lang, ".@")
```
lslezak: You can remove the optional `.` or `@` suffix by a single call:
```ruby
local_lang.sub!(/[.@].*$/, "")
```

language/src/modules/Language.rb
```diff
@@ -586,6 +600,21 @@ def ReadSysconfigValues
...
 
+    # read UTF-8 status from localed.conf
+    def ReadUtf8Setting
+      return nil if Mode.testsuite
+      # during live installation, we have local configuration
+      if !Stage.initial || Mode.live_installation
+        @localed_conf  = Y2Country.read_locale_conf
+        return nil if @localed_conf.nil?
+        local_lang = @localed_conf["LANG"]
+        @use_utf8 = Builtins.search(local_lang, ".UTF-8") != nil unless local_lang.empty?
```
lslezak: Simplification:
```ruby
local_lang.include?(".UTF-8")
```
But additionally you might need to check for `nil`...


IIRC you can use also the downcase variant and possibly without the dash, e.g. `.utf8`. Should we handle that as well? Command `locale -a` prints only the `.utf8` variant...

package/yast2-country.spec
```diff
@@ -40,6 +40,7 @@ BuildRequires:  yast2-core >= 3.1.12
 BuildRequires:  yast2-ruby-bindings >= 3.1.26
 # OSRelease.id
 BuildRequires:  yast2 >= 3.2.9
+BuildRequires:  rubygem(%{rb_default_ruby_abi}:ruby-dbus)
```
lslezak: Do we really need that for building the package? The `read_locale_conf` call is mocked in the tests and I'm not sure if DBus is actually running in the build chroot... And using DBus during package build is strange, I'd avoid this dependency if possible.

language/src/lib/y2country/language_dbus.rb
```diff
@@ -0,0 +1,63 @@
...
+        # inst-sys (because of constructor)
+        log.info("DBus module not available")
+        return nil
+      end
+      localed_conf = {}
+      # https://www.freedesktop.org/wiki/Software/systemd/localed/
+      sysbus = DBus.system_bus
+      locale_service   = sysbus["org.freedesktop.locale1"]
+      locale_object    = locale_service.object "/org/freedesktop/locale1"
+      locale_object.introspect # needed, ask mvidner for explanation
```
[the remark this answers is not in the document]
mvidner: Right! https://github.com/mvidner/ruby-dbus/issues/28.
```rb
# needed, https://github.com/mvidner/ruby-dbus/issues/28
```
But 8 days ago a newer version of ruby-dbus was accepted to SLE-15, so we will be able to remove this if we really want to.

package/yast2-country.spec
```diff
@@ -40,6 +40,7 @@ BuildRequires:  yast2-core >= 3.1.12
 BuildRequires:  yast2-ruby-bindings >= 3.1.26
 # OSRelease.id
 BuildRequires:  yast2 >= 3.2.9
+# BuildRequires:  rubygem(%{rb_default_ruby_abi}:ruby-dbus)
```
lslezak: If it is not needed then remove it completely, no dead code please...

language/src/lib/y2country/language_dbus.rb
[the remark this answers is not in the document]
lslezak: Great! So remove I'd remove it if possible. Then maybe we should `Require` the fixed version (`>= 0.13.0`).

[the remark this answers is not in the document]
jsrain: I'd like to keep it, some of us still develop on older openSUSE version from git (therefore no versioned RPM dependency) and this could turn into a huge pain point trying to find out why it does not work.

PR #153, opened 2017-12-04
do not touch RC_* variables in sysconfig/language (they still keep in use, limiting the meaning to default shell (fate#319454)

jsrain:


mvidner:
What about this, which is still in?
https://github.com/jsrain/yast-country/blob/b1295adb2c58b5912291f8f41b9de7db13b14a92/console/src/modules/Console.rb#L200

package/yast2-country.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Dec  4 11:30:50 UTC 2017 - jsrain@suse.cz
+
+- do not touch RC_* variables in sysconfig/language (they still
+  keep in use, limiting the meaning to default shell (fate#319454)
```
mvidner: This is hard to understand. What is the impact for the user/admin? Can we name all the vars (RC_LANG only?) instead of using a wild card?

jsrain:
Valid comment (Console adaptation is pending). Code adjusted.

PR #155, opened 2017-12-13
AY writing keyboard settings

schubi2:


keyboard/test/keyboard_test.rb
```diff
@@ -462,6 +485,14 @@ module Yast
           end
         end
       end
+
+      context "AutoYaST upgrade mode" do
+        it "do not make a new proposal" do
```
imobachgs: does not

package/yast2-country.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Dec 13 13:48:06 CET 2017 - schubi@suse.de
+
+- AutoYaST upgrade fix: Writing keyboard settings defined in AY
```
imobachgs: np: I would change "AY configuration file" for "AutoYaST profile"

PR #158, opened 2017-12-14
Further adapatations for locale settings

jsrain:


console/src/modules/Console.rb
```diff
@@ -130,9 +130,11 @@ def SelectFont(lang)
     # save data to system (rc.config agent)
 
     def Save
-      SCR.Write(path(".sysconfig.console.CONSOLE_FONT"), @font)
-      SCR.Write(path(".sysconfig.console.CONSOLE_SCREENMAP"), @screenMap)
-      SCR.Write(path(".sysconfig.console.CONSOLE_UNICODEMAP"), @unicodeMap)
+      SCR.Write(path(".etc.vconsole_conf.FONT"), @font)
```
lslezak: Is it OK to write these values directly? For KEYMAP we were told to not write it directly by YaST, see https://bugzilla.suse.com/show_bug.cgi?id=1023774.

Is it OK for the FONT values?

jsrain: Yes, it is at least for now since there is no other interface at the moment. Might change in the future.

PR #159, opened 2018-01-08
adjust calling systemd-firstboot for keyboard setting (bsc#1074481)

jsrain:


keyboard/src/modules/Keyboard.rb
```diff
@@ -887,19 +887,26 @@ def Save
...
         # this sets both the console and the X11 keyboard (see "man localectl")
         "/usr/bin/localectl set-keymap #{chomped_keymap}"
       end
       log.info "Making keyboard setting persistent: #{cmd}"
-      result = SCR.Execute(path(".target.bash_output"), cmd)
+      result = if Stage.initial
+        WFM.Execute(path(".local.bash_output"), cmd)
+      else
+        SCR.Execute(path(".target.bash_output"), cmd)
+      end
```
jreidinger: well, to be honest I prefer here if both part will be in separated branch, as it makes it easier to read like:
```
if Stage.initial
  # do use --root option, running in chroot does not work (bsc#1074481)
  cmd = "/usr/bin/systemd-firstboot --root '#{Installation.destdir}' --keymap '#{chomped_keymap}'"
  result = WFM.Execute(path(".local.bash_output"), cmd)
else
  # this sets both the console and the X11 keyboard (see "man localectl")
  cmd = "/usr/bin/localectl set-keymap #{chomped_keymap}"
  result = SCR.Execute(path(".target.bash_output"), cmd)
end

log.info "command #{cmd} end with #{result.inspect}"
```

PR #161, opened 2018-01-24
Use eurlatgr font instead of lat9w-16

imobachgs:
Replaces `lat9w-16` with `eurlatgr`.


Vogtinator:
Isn't it possible to just merge consolefonts with consolefonts_opensuse.ycp?

imobachgs:
@Vogtinator Yes, we could merge both of them. But we would need to fix some conflicts. Basically:

* If you check `de_DE`, `de_DE`, `de_CH` and `de` for SLE, all of them are using `lat1-16.psfu`. We would keep them, but `de` is also used as fallback in the openSUSE file, so we have a conflict here (not only the font, but the unicode map, screen map and console magic parameters).

On the other hand, should we keep entries like `en_US.UTF-8`? openSUSE would fallback to `en`, which contains a different set of values for unicode map, screen map and console magic.

Opinions?

Vogtinator:
IMO all of those should use eurlatgr, like the opensuse map says.
Then you shouldn't need any unicode or screen maps anymore.

imobachgs:
Ok, then switching to use the openSUSE map, as it is, should do the trick. I only have spotted one conflict using this approach: `tr` used `iso09.f16n` in SLES but `eurlatgr` in openSUSE. I guess it is safe to replace it with `eurlatgr`.

Vogtinator:
Looks good to me - I searched for `iso09.f16n` everywhere but the only hits were in YaST or related parts so AFAICT it should be safe to not use it for `tr`.

imobachgs:
@Vogtinator Thanks. I will merge as soon as Travis is happy again.

package/yast2-country.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jan 24 13:39:34 UTC 2018 - igonzalezsosa@suse.com
+
+- Use eurlatgr font instead of lat9w-16 (bsc#1068814).
```
jreidinger: maybe also mention that with this change we sync opensuse and SLE

PR #166, opened 2018-02-14
Add simple language selection

imobachgs:
Add a widget which allows selecting a language from a given list. The main difference with `Y2Country::Widgets::LanguageSelection` is that it does not make any change to the system (like changing the Language).

PBI: https://trello.com/c/VjEpcE8x/


imobachgs:
I am closing this PR because this widget will be included in `yast2-packager` to avoid circular dependencies. If needed, it can be moved to yast2.rpm in the future.

PR #169, opened 2018-05-14
handle setting locale in firstboot (bsc#1092979)

jsrain:


jsrain:
agraf confirmed that the patch (as sent per IRC on Friday) fixed the problem

PR #171, opened 2018-05-15
Ntp pool misusage

jreidinger:
trello: 
bzilla: https://bugzilla.suse.com/show_bug.cgi?id=1090168

tested manually

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -575,7 +575,7 @@ module Yast
         # configure NTP client
         # to prevent misusage of ntp.org we need to distinguish opensuse and SLE usage
         base_products = Product.FindBaseProducts
-        if base_products.any? { |p| p["shortname"] == "openSUSE" }
+        if base_products.any? { |p| p["name"] =~ /openSUSE/i }
           servers = (0..3).map { |i| "#{i}.opensuse.pool.ntp.org" }
         else
           servers = (0..3).map { |i| "#{i}.novell.pool.ntp.org" }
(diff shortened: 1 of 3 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -572,22 +573,21 @@ def TimezoneDialog(args)
         # true by default (fate#303520)
         @ntp_used = true
         # configure NTP client
-        Builtins.srandom
-        @ntp_server = Builtins.sformat(
-          "%1.opensuse.pool.ntp.org",
-          Builtins.random(4)
-        )
+        # to prevent misusage of ntp.org we need to distinguish opensuse and SLE usage
+        base_products = Product.FindBaseProducts
```
lslezak: Hm, this `Product` module initializes libzypp which might take some time (refreshing the services or repositories). 

What about using the simpler [OSRelease](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/OSRelease.rb) module instead? That just parses the `/etc/os-release` without any overhead.
And that file is also present in the inst-sys so it will work correctly also during the installation.

lslezak: I see, if it is used only in the installation then it's OK.

timezone/src/include/timezone/dialogs.rb
```diff
@@ -572,22 +573,21 @@ def TimezoneDialog(args)
...
+        if base_products.any? { |p| p["shortname"] == "openSUSE" }
```
lslezak: Here I'd prefer just `p["name"]` as that is the unique ID for the product, `shortname` or `displayname` are long texts which might contain anything displayable to the user, e.g. the `shortname`  might be changed to "openSUSE-16" or something like that in the future.

Maybe to make the test more robust we could make it less strict, e.g. `p["name"] =~ /openSUSE/i`

PR #177, opened 2018-07-24
Pass absolute paths to "yard"

lslezak:
For `yast2-country` we need a specific fix, the generic fix in `yast2-devtools` (https://github.com/yast/yast-devtools/pull/130) does not work here because of a different directory layout (the sources are in several subdirectories).

See https://github.com/yast/yast-devtools/pull/130 for more details.


lslezak:
Travis fails, for SP1 it uses the old Ubuntu way which is broken now, let's ignore that...

PR #182, opened 2018-09-27
Set only system date if no HW clock available

shundhammer:
## Trello

https://trello.com/c/Hxd4d3qV/301-sles12-sp4-p2-1103744-rpi-unable-to-set-time-manually-in-firstboot-no-rtc-hw

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1103744

## Problem

If a hardware clock is not available (for example on a RasPi), setting the date and time during installation has no effect: It tries to set it via the `hwclock` command. If that fails, it does not set it at all.

## Solution

If that `hwclock` command for setting date and time fails (and only then), use the good old `date` command to set at least the system clock.

This is done independently of RasPi or whatever hardware; it might as well fail on other hardware.

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -472,51 +472,70 @@ module Yast
 
     # Set the new time and date given by user
     def SetTime(year, month, day, hour, minute, second)
-      if !Arch.s390
-        date = Builtins.sformat(
-          " --date=\"%1/%2/%3 %4:%5:%6\" ",
-          month,
-          day,
-          year,
-          hour,
-          minute,
-          second
-        )
-        cmd = ""
-        if Ops.greater_than(Builtins.size(@timezone), 0) &&
-            @hwclock != "--localtime"
-          cmd = Ops.add(Ops.add("TZ=", @timezone), " ")
-        end
-        cmd = Ops.add(
-          Ops.add(Ops.add(cmd, "/sbin/hwclock --set "), @hwclock),
-          date
-        )
-        Builtins.y2milestone("SetTime cmd %1", cmd)
-        SCR.Execute(path(".target.bash"), cmd)
-        cmd = Ops.add("/sbin/hwclock --hctosys ", @hwclock)
-        Builtins.y2milestone("SetTime cmd %1", cmd)
-        SCR.Execute(path(".target.bash"), cmd)
-        # actually, it was probably not called, but do not let it change the time again after manual change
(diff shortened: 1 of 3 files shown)


timezone/src/modules/Timezone.rb
```diff
@@ -634,27 +649,23 @@ def GetDateTime(real_time, locale_format)
             )
             ds = Ops.add(ds, Ops.multiply(t, 60))
             ds = Ops.unary_minus(ds) if Builtins.substring(tzd, 0, 1) == "-"
-            Builtins.y2milestone("GetDateTime ds %1 diff %2", ds, @diff)
+            log.info("GetDateTime ds: #{ds} diff: #{diff}")
```
shundhammer: And another typo: @diff

timezone/src/modules/Timezone.rb
```diff
@@ -613,19 +632,15 @@ def GetDateTime(real_time, locale_format)
...
+      log.info("GetDateTime hwclock: #{@hwclock} real: #{real_time}")
       if !real_time && !Mode.config
         ds = 0
         if @diff != 0
           out2 = Convert.to_map(
             SCR.Execute(path(".target.bash_output"), "date +%z")
           )
           tzd = Ops.get_string(out2, "stdout", "")
-          Builtins.y2milestone("GetDateTime tcd=%1", tzd)
+          log.info("GetDateTime tcd: #{tzd}")
```
shundhammer: Duh - typo. Will fix.

timezone/src/modules/Timezone.rb
```diff
@@ -472,53 +472,72 @@ def CallMkinitrd
...
     end
 
+    # Set the hardware clock with the given date.
+    # @param timedate [String]
+    # @return [Bool] true if success, false if error
+    #
+    def set_hwclock(date)
+      cmd = tz_prefix + "/sbin/hwclock --set #{@hwclock} --date=\"#{date}\""
+      log.info("set_hwclock: #{cmd}")
+      SCR.Execute(path(".target.bash_output"), cmd) == 0
```
lslezak: The `.target.bash_output` call returns a `Hash`, looks like the `.target.bash` should be used here instead...

shundhammer: You are right.

For a while I had considered checking the error message of that command for that specific text, but then I decided against it and to just always try to set the system time when `hwclock` fails to set it via the RTC.

PR #186, opened 2018-11-11
Installation ntp first run: Write and sync the default config if enabled in the control file

teclator:
Description
--------------

In this [PR](https://github.com/yast/yast-ntp-client/pull/121), we avoided to write when the client was calling for syncing only.

So this PR is for removing the sync only parameter for also writing during the first run of the dialog in an installation.

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -47,6 +47,7 @@ module Yast
       Yast.import "NetworkService"
       Yast.import "Package"
       Yast.import "Popup"
+      Yast.import "Product"
       Yast.import "ProductFeatures"
       Yast.import "Service"
       Yast.import "Stage"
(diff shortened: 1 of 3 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -573,21 +573,18 @@ def TimezoneDialog(args)
         # true by default (fate#303520)
         @ntp_used = true
         # configure NTP client
-        Builtins.srandom
-        @ntp_server = Builtins.sformat(
-          "%1.opensuse.pool.ntp.org",
-          Builtins.random(4)
-        )
+        # to prevent misusage of ntp.org we need to distinguish opensuse and SLE usage
+        base_products = Product.FindBaseProducts
```
lslezak: There is a missing `Yast.import "Product"` at the top.

dgdavid: Good catch!

PR #190, opened 2018-11-12
Timezone: Offer dhcp servers provided by DHCP (fate#323454)

teclator:
- [Trello Card](https://trello.com/c/LWJH8yhO/403-sle-15-mu-ntp-epic-f323454-chrony-offer-to-use-dhcp-provided-list-of-ntp-servers)

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -581,6 +581,9 @@ module Yast
           servers = (0..3).map { |i| "#{i}.novell.pool.ntp.org" }
         end
         @ntp_server = servers.sample
+        # Dot not select a dhcp ntp server by default but add it to the offered
+        # list (fate#323454)
+        servers = ntp_call("dhcp_ntp_servers", {}).concat(servers).uniq
         argmap = {
           "server"       => @ntp_server,
           # FIXME ntp-client_proposal doesn't understand 'servers' yet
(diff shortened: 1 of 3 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -581,6 +581,9 @@ def TimezoneDialog(args)
           servers = (0..3).map { |i| "#{i}.novell.pool.ntp.org" }
         end
         @ntp_server = servers.sample
+        # Dot not select a dhcp ntp server by default but add it to the offered
+        # list (fate#323454)
+        servers = ntp_call("dhcp_ntp_servers", {}).concat(servers).uniq
```
lslezak: I'm a bit confused, why we cannot simply call `Yast::Lan.dhcp_ntp_servers` here directly and need to go via calling the `ntp-client_proposal` client which calls `NtpClient.dhcp_ntp_servers` which calls `Yast::Lan.dhcp_ntp_servers`... :question:

teclator: Yeah, sounded quite complicated. But basically, I avoided to add an extra dependency here. I mean, we avoided to import or require ntp-client, so for having the same behavior I exposed it trough ntp-client_proposal.

There I originally called Yast::Lan.dhcp_ntp_servers directly, but as I needed the method in different places having to import Yast::Lan decided to add a convenience method directly in NtpClient.

lslezak: Actually it would add the `yast2-network` package dependency, I'm not if that would cause some cyclic dependency... It would require some deeper analysis, so to be on the safe side I'm fine with this indirect solution.

PR #193, opened 2018-11-20
Do not hide manual checkbox in 80x25

teclator:
- Needed by yast/yast-ntp-client#125

screenshot from 2018-11-20 09-21-20
screenshot from 2018-11-20 09-21-17

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -302,6 +302,7 @@ module Yast
 
 
       cont = VBox(
+          VSpacing(UI.TextMode ? 1 : 0),
           HBox(
           HWeight(1, VBox()),
           HWeight(
@@ -316,7 +317,7 @@ module Yast
                   HSpacing(3),
                   VBox(
                     Left(timeterm),
-                    VSpacing(),
+                    UI.TextMode ? Empty() : VSpacing(),
                     Left(dateterm),
                     VSpacing(),
                     HBox(
@@ -333,7 +334,7 @@ module Yast
                     )
                   )
                 ),
-                VSpacing(1),
+                VSpacing(UI.TextMode ? 0.5 : 1),
                 Left(
                   RadioButton(
                     Id(:ntp),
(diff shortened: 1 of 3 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -333,7 +334,7 @@ def SetTimeDialog
                     )
                   )
                 ),
-                VSpacing(1),
+                VSpacing(UI.TextMode ? 0.5 : 1),
```
lslezak: Using a floating point in text mode technically does not make sense, either use `VSpacing(UI.TextMode ? 0 : 1)` or `VSpacing(0.5)`. Or did I overlook something?

timezone/src/include/timezone/dialogs.rb
```diff
@@ -302,6 +302,7 @@ def SetTimeDialog
 
 
       cont = VBox(
+          VSpacing(UI.TextMode ? 1 : 0),
```
lslezak: Is it correct? Usually we add more space in graphical mode, not in text mode...

timezone/src/include/timezone/dialogs.rb
```diff
@@ -316,7 +317,7 @@ def SetTimeDialog
                   HSpacing(3),
                   VBox(
                     Left(timeterm),
-                    VSpacing(),
+                    UI.TextMode ? Empty() : VSpacing(),
```
lslezak: Why not `VSpacing(UI.TextMode ? 0 : 1)`? That does not work here?

timezone/src/include/timezone/dialogs.rb
teclator: I added this extra VSpace at the top but removed at the end in textmode removing the space between the buttons and the Header, instead of this trick which other option do you recommend?

timezone/src/include/timezone/dialogs.rb
teclator: Just was not sure about VSpacing default, so to left VSpacing() instead of forcing a `1`

timezone/src/include/timezone/dialogs.rb
teclator: No, you didn't, you are completely right.

timezone/src/include/timezone/dialogs.rb
lslezak: I'm fine with that, I just wanted to double check as usually it's the other way round.

PR #196, opened 2018-11-22
Keyboard rewrite: select current layout on startup

jyeray:

In this pull request the current keyboard layout in the system is selected when start the module.


schubi2:

PR #197, opened 2018-11-22
 Keyboard rewrite: add more layouts to the module

jyeray:

Layouts from old module has been added. Also now its possible to have two differents keyboard layouts with same code, "us" for example, but different description for the user.


schubi2:

PR #198, opened 2018-11-22
 Keyboard rewrite: fix error setting x11 keymap

jyeray:

An error while setting the x11 keymap is fixed. It happened because the parameters passed to setxkbmap command was invalid in some cases.


schubi2:

PR #200, opened 2018-11-22
Keyboard rewrite: encapsulate systemd actions

jyeray:

Some responsibilities has been split in different classes, in order to keep the systemd logic encapsulated.


schubi2:

jyeray:
Thanks @schubi2, that are great news! I am happy to read that words. I hope to see it in Tumbleweed soon 😄

PR #203, opened 2019-02-08
fix timezone setting when switching utc/local time (bsc#1087228)

wfeldt:
The timezone code is really convoluted stuff bordering on bizarr. It deserves some proper refactoring.

Nevertheless this patch solves the reported issue. (And others I encountered during testing.)

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -877,6 +877,20 @@ module Yast
           end
         end
 
+        if ret == :hwclock
+          # Keep internal state in sync (Timezone.hwclock <-> @hwclock_s) in
+          # case user enters SetTimeDialog() and sets the time (bsc#1087228).
+
+          @hwclock_s = UI.QueryWidget(Id(:hwclock), :Value) ? :hwclock_utc : :hwclock_localtime
+          SetTimezone(@hwclock_s, timezone, false, false)
+
+          Builtins.y2milestone("hwclock changed to: %1 (%2), diff: %3", @hwclock_s, Timezone.hwclock, Timezone.diff)
+
+          # restart input loop
+          ret = :again
+          next
+        end
+
         if ret == :region
           num = selected_region.call
           next if num == sel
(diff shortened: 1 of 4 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -877,6 +877,20 @@ def TimezoneDialog(args)
...
         end
 
+        if ret == :hwclock
+          # Keep internal state in sync (Timezone.hwclock <-> @hwclock_s) in
+          # case user enters SetTimeDialog() and sets the time (bsc#1087228).
+
+          @hwclock_s = UI.QueryWidget(Id(:hwclock), :Value) ? :hwclock_utc : :hwclock_localtime
+          SetTimezone(@hwclock_s, timezone, false, false)
+
+          Builtins.y2milestone("hwclock changed to: %1 (%2), diff: %3", @hwclock_s, Timezone.hwclock, Timezone.diff)
```
lslezak: NP: maybe I'd switch to the new logging with `log.info` (do not forget to `include Yast::Logger`)

timezone/src/include/timezone/dialogs.rb
```diff
@@ -954,7 +970,8 @@ def TimezoneDialog(args)
             ret = :again
             timezone = timezone_old
           end
-          Builtins.y2milestone("timezone %1 ret %2", timezone, ret)
+
+          Builtins.y2milestone("timezone %1 ret %2, hwclock %3 -> %4", timezone, ret, hwclock_s_old, @hwclock_s)
```
lslezak: NP: the same here

wfeldt: As said above the code is rather in need of some refactoring effort. It's not only the ycpkiller leftovers but also the whole code organization with the internal state distributed everywhere,
unused method arguments, functionless dialog elements, and misleading button labels.

So I'd rather not change the logging interface at this point.

wfeldt: Maybe I should add a doc listing the issues I saw.

dgdavid: > Maybe I should add a doc listing the issues I saw.

It seems to me a nice idea :)

timezone_todo.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Change Date and Time dialog
+
+- In `Manually` there is a `Change the Time Now` checkbox. If you uncheck
+  it the whole `Manually` item is grayed out - so what's the point?
+  AFAICS this checkbox could be removed.
+
+## Integration testing
+
+When doing integration tests, verify that:
```
lslezak: This *how to test* docu is great! :+1:

PR #205, opened 2019-03-05
Stop using obsolete XVersion

mvidner:
It was useful in 2006 when X.org 6 and X.org 7 had different places for files

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -121,7 +121,6 @@ module Yast
       Yast.import "OSRelease"
       Yast.import "ProductFeatures"
       Yast.import "Stage"
-      Yast.import "XVersion"
       Yast.import "Report"
 
       # ------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -559,7 +558,7 @@ def SetKeyboard(keyboard)
       # do not try to run this with remote X display
       if Ops.greater_than(Builtins.size(@Apply), 0) && x11_setup_needed
         # Apply cannot be escaped as it is already set of parameters. But it is at least our string and not user provided.
-        @xkb_cmd = "#{File.join(XVersion.binPath, "setxkbmap")} #{@Apply}"
+        @xkb_cmd = "/usr/bin/setxkbmap #{@Apply}"
```
lslezak: Is `@Apply` safe or should we `.shellescape` it?

jreidinger: See comment above code I wrote when I do escaping ;)

lslezak: Ah,...:dizzy_face:

mvidner:
Ugh, a gh reference is not enough for Jenkins. I've created https://bugzilla.suse.com/show_bug.cgi?id=1144627 and will fill that in

PR #206, opened 2019-03-14
Set language as read-only

dgdavid:
## Problem

During Kubic installation, it is allowed to configure a different locale. However, by default, only kubic-locale-archive is installed (not glibc-locale), which means that only `en` and POSIX locales will be available in the installed system.

So, the selected language during installation **should only be used for the installer**, the system itself should always be en_US.UTF-8.

- https://bugzilla.suse.com/show_bug.cgi?id=1121256
- https://trello.com/c/kww2Wsx0/807-5-kubic-1121256-kubic-installs-with-invalid-locale-setup

## Solution

Analogous to how [it was done for the Timezone](https://github.com/yast/yast-country/pull/97), a `readonly_language` feature will checked to decide if persist the chosen language during the installation or use the [`DEFAULT_FALLBACK_LANGUAGE`](https://github.com/yast/yast-country/blob/2a3fb6e460b4eec78d8e3c8dc627b07bdd780f3f/language/src/modules/Language.rb#L36) instead.

## Improvements

- Language#Save [refactored](https://github.com/yast/yast-country/pull/206/commits/38b05d3d0fcd24c4e5da93c6a9e658aa1ad9907c)

  Among other changes, now it uses [`Yast::Execute.locally!`](https://github.com/yast/yast-yast2/blob/62f62b02018caa3224f5f880aa62782e62554706/library/system/src/lib/yast2/execute.rb#L107) instead `Yast::WFM` or `Yast::SCR`.

## Testing

- Added necessary test examples
- Adapted necessary unit tests
- Also tested manually

- :warning: **Removed the old testsuite for Language#Save**, because it is already tested with RSpec (besides that, it started failed with the change from `Yast::(WFM|SCR).Execute` to `Yast::Execute.locally!` described above)

## Related PRs

* https://github.com/yast/yast-installation-control/pull/78
* https://github.com/yast/yast-installation/pull/789
* https://github.com/yast/skelcd-control-Kubic/pull/32

## Additional notes

Better to review commit by commit.


language/test/Language_test.rb
```diff
@@ -162,6 +164,86 @@
...
+        subject.SetDefault
+    end
+
+    it "writes .sysconfig.language.INSTALLED_LANGUAGES" do
+      expect(Yast::SCR).to receive(:Write).with(Yast.path(".sysconfig.language.INSTALLED_LANGUAGES"), anything)
+
+      subject.Save
+    end
+
+    it "cleans .sysconfig.language in sysconfig" do
```
jreidinger: this is wrong. It does not mean cleaning. It is pushing cache. If you write to file path in SCR nil, it means flush cache and write immediately.

language/test/Language_test.rb
```diff
@@ -162,6 +164,86 @@
...
+        allow(Yast::WFM).to receive(:Execute).and_return(command_result)
+        allow(Yast::SCR).to receive(:Execute).and_return(command_result)
+        allow(Yast::SCR).to receive(:Write)
+        allow(subject).to receive(:valid_language?).with(language).and_return(true)
+
+        subject.Set(language)
+        subject.SetDefault
+    end
+
+    it "writes .sysconfig.language.INSTALLED_LANGUAGES" do
```
jreidinger: this is a bit low level to my taste. What actually you do it that you modify /etc/sysconfig/language value INSTALLED_LANGUAGES. Ideally tested with switched SCR on real file, that is really written down, but it is just nice to have. But better description what method do is almost must have.

language/test/Language_test.rb
```diff
@@ -162,6 +164,86 @@
...
+      let(:language) { "zh_HK" }
+
+      it "sets LC_MESSAGES to zh_TW" do
+        expect(Yast::SCR).to receive(:Execute).with(anything, /.*localectl set-locale.*LC_MESSAGES.*zh_TW.*/)
+
+        subject.Save
+      end
+    end
+
+    context "when LC_MESAGES is zh_TW" do
```
jreidinger: typo MESAGES

language/test/Language_test.rb
```diff
@@ -162,6 +164,86 @@
...
+      before do
+        allow(subject).to receive(:@localed_conf).and_return({ "LC_MESSAGES" => "zh_TW" })
+      end
+
+      context "and language is not zh_HK" do
+        it "cleans LC_MESSAGES" do
+          expect(Yast::SCR).to_not receive(:Execute).with(anything, /.*localectl.*LC_MESSAGES.*zh_TW.*/)
```
jreidinger: you test that it is not called, but it looks different then cleaning it for me. Does it really write empty string there?

language/src/modules/Language.rb
```diff
@@ -915,53 +916,20 @@ def Selection
...
-        log.error "output: #{result.inspect}"
-        # TRANSLATORS: the "%s" is replaced by the executed command
-        Report.Error(_("Could not save the language setting, the command\n%s\nfailed.") % cmd)
-      else
-        log.info "output: #{result.inspect}"
-      end
+      log.info("Making language settings persistent: #{command}")
 
-      Builtins.y2milestone("Saved data for language: <%1>", loc)
+      Yast::Execute.locally!(command.split)
```
jreidinger: sorry, this cannot work this way. Using split is very bad idea. We shellescape arguments, but it can be something like `British English` which we escape to `British\ English` and this split break it. Also Cheetah does not use shell. So shellescape here will break it heavily.

language/src/modules/Language.rb
```diff
@@ -915,53 +916,20 @@ def Selection
...
+      nil
+    rescue Cheetah::ExecutionFailed => e
+      log.error "Language configuration not written.\n#{e.inspect}"
 
+      # TRANSLATORS: the "%s" is replaced by the executed command
+      Report.Error(_("Could not save the language setting, the command\n%s\nfailed.") % command)
```
jreidinger: please try to write also stderr to give user hint why it failed. Not just what.

language/src/modules/Language.rb
```diff
@@ -1437,12 +1405,70 @@ def readonly
...
+    # It is different depending on the stage, since for a not installed system it is needed to use
+    # the `systemd-firsrboot` tool, which does not work in a chroot
+    #
+    # @param locale [String]
+    #
+    # @return [String] the command to be executed
+    def save_command_for(locale)
+      if Stage.initial
+        # do use --root option, running in chroot does not work
+        "/usr/bin/systemd-firstboot --root #{Installation.destdir.shellescape} --locale #{locale.shellescape}"
```
jreidinger: if you want to use it for cheetah, simply use array like `["/usr/bin/systemd-firstboot", "--root", Installation.destdir, "--locale",  locale]`

language/test/Language_test.rb
```diff
@@ -202,7 +200,7 @@
       let(:language) { "zh_HK" }
 
       it "sets LC_MESSAGES to zh_TW" do
-        expect(Yast::SCR).to receive(:Execute).with(anything, /.*localectl set-locale.*LC_MESSAGES.*zh_TW.*/)
+        expect(Yast::Execute).to receive(:locally!).with(array_including(/LC_MESSAGES\\=zh_TW/))
```
jreidinger: this `\\` looks wrong to me.

language/test/Language_test.rb
```diff
@@ -46,6 +46,8 @@
     allow(Y2Country).to receive(:read_locale_conf).and_return(nil)
     allow(subject).to receive(:languages_map).and_return(languages_map)
     allow(subject).to receive(:GetLanguagesMap).and_return(languages_map)
+
+    Yast::Language.main
```
jreidinger: this change will depends on fact that main will reset all values. Is it really true?

language/src/modules/Language.rb
```diff
@@ -1402,7 +1402,8 @@ def fbiterm?
     #
     # @return [Boolean] true if it's read-only; false otherwise.
     def readonly
-      ProductFeatures.GetBooleanFeature("globals", "readonly_language")
+      @readonly unless @readonly.nil?
+      @readonly = ProductFeatures.GetBooleanFeature("globals", "readonly_language")
```
jreidinger: is it needed? How big impact does it have? BTW this is not reset in main, so exactly problem with dirty value I point in previous commit.

language/test/Language_test.rb
dgdavid: I think so, but it is possible that I overlooked something. Let me recheck it.

language/src/modules/Language.rb
dgdavid: > is it needed? How big impact does it have?

Only for consistency with the same changes in the Timezone module

> BTW this is not reset in main

Arrrg! I forgot to add the `@readonly = nil` statement in #main.

language/src/modules/Language.rb
dgdavid: I think that the "stderr" do not give to the user (I mean, normal user) too much information. User who find the "stderr" useful probably will look for it in the logs, isn't it?

That said, I did not touch this intentionally because the "text freeze". So, for now, I prefer to keep it as it is/was.

language/src/modules/Language.rb
dgdavid: Thank you Josef, I though that it will be the same, but your previous clarification allowed me to realize that it is not.

language/test/Language_test.rb
dgdavid: If I didn't check it wrong, it seems that all the available instances variables in the module are being properly initialized in the `#main` . Namely, 

```ruby
@available_lang_filenames
@cjk_languages
@default_language
@english_names
@ExpertSettingsChanged
@lang2keyboard
@lang2timezone
@language
@language_on_entry
@languages
@languages_directory
@languages_map
@languages_on_entry
@linuxrc_language_set
@localed_conf
@locales
@name
@preselected
@readonly [1]
@reset_recommended
@secondary_items
@selection_skipped
@text_mode
@translation_status
@use_utf8
```

[1] which I'll fix in next commit.

language/src/modules/Language.rb
jreidinger: ok, text freeze is good point. There are users which understand stderr, but do not look immediatelly to logs. And it also really helps in openqa as you see it in screenshot.

language/test/Language_test.rb
dgdavid: If it is cleaned (a.k.a [deleted]()), what I expect if that the 

> /usr/sbin/localectl set-locale

command does not receive the `LC_MESSAGE=zh_TW` as an argument. 

Is that approach wrong?

language/src/modules/Language.rb
```diff
@@ -1402,6 +1402,25 @@ def fbiterm?
...
 
+    # Returns the command to make language settings persistent
+    #
+    # It is different depending on the stage, since for a not installed system it is needed to use
+    # the `systemd-firsrboot` tool, which does not work in a chroot
+    #
+    # @param locale [String]
+    #
+    # @return [Array] an array containing the command to be executed
+    def save_command_for(locale)
```
jreidinger: NP: I am not sure if method name is good. IT is a bit misleading for me. Maybe something like `store_locale_command` sounds better for me. But just np.

dgdavid: Changed ;)

language/src/modules/Language.rb
```diff
@@ -1431,6 +1402,79 @@ def fbiterm?
...
+    # It is different depending on the stage, since for a not installed system it is needed to use
+    # the `systemd-firsrboot` tool, which does not work in a chroot
+    #
+    # @param locale [String]
+    #
+    # @return [Array<String>] an array containing the command to be executed
+    def store_locale_command(locale)
+      if Stage.initial
+        # do use --root option, running in chroot does not work
+        ["/usr/bin/systemd-firstboot", "-root", Installation.destdir, "--locale", locale]
```
DimStar77: Invalid command parameter - in the past this was ```--root``` (not only ```-root```)

Very likely this is also the cause of e.g. https://openqa.opensuse.org/tests/883387#step/await_install/21

dgdavid: Oops! 

It is a mistake made [during an update from code review](https://github.com/yast/yast-country/pull/206/commits/104c69cb123c71e1ce05046688f8a4f294a44b7a#diff-3b96e2c564f7d2e272c9457029f49ad5L1465). My fault :sweat_smile: I'll fix it ASAP.

Thank you!

DimStar77: Already fixed and merged with https://github.com/yast/yast-country/pull/207

dgdavid: I just see it :)

So, thank you so much! :)

PR #209, opened 2019-03-28
Fix in setting console keyboard

schubi2:
The sulution is more or less described in:
https://bugzilla.suse.com/show_bug.cgi?id=1120957#c16

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -177,6 +177,10 @@ module Yast
       #
       @ckb_cmd = ""
 
+      # The seriell console keyboard command
+      #
+      @skb_cmd = ""
+
       # The X11 keyboard command
       #
       @xkb_cmd = ""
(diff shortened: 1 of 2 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -177,6 +177,10 @@ def main
       #
       @ckb_cmd = ""
 
+      # The seriell console keyboard command
```
imobachgs: *seriell*? Did you mean *serial*?

keyboard/src/modules/Keyboard.rb
```diff
@@ -992,11 +997,17 @@ def SetConsole(keyboard)
 
         Builtins.y2milestone("Setting console keyboard to: <%1>", @current_kbd)
         Builtins.y2milestone("loadkeys command: <%1>", @ckb_cmd)
-
         SCR.Execute(path(".target.bash"), @ckb_cmd)
+
+        # It could be that for seriell tty's the keyboard cannot be set. So it will
+        # be done seperately in order to ensure that setting console keyboard
```
imobachgs: separately

PR #217, opened 2019-06-17
Port to rake

hellcp:


lslezak:
Hm, after removing the `**/testsuite/Makefile.am` files all the old tests in `**/testsuite/tests` become unused dead files. :worried: 

I'm not sure how good the old tests are but I'm a bit worried about loosing them all...

jreidinger:
replaced by https://github.com/yast/yast-country/pull/253

PR #218, opened 2019-07-22
Fix opensuse keymaps

ancorgs:
This is a previous step for unifying SLE and openSUSE keymaps to use the ones from `keyboard_raw_opensuse.ycp`. While checking the current status we found some errors in the openSUSE file and I would like to fix them first (with its own version and so on) before performing the unification.

To get the full picture, read [bsc#1124921](https://bugzilla.suse.com/show_bug.cgi?id=1124921) starting at comment 18 (the previous comments refer to issues fixed a long time ago).

Part of https://trello.com/c/vYVMPFWZ/971-follow-up-of-1124921-try-to-merge-keyboardrawopensuseycp and also partially related to https://trello.com/c/znCthGFV/1142-5-sle15-sp1-p2-1140872-sles15-sp1-systemd-vconsole-setupservice-failed

### DIFF
--- a/keyboard/src/data/keyboard_raw_opensuse.ycp
+++ b/keyboard/src/data/keyboard_raw_opensuse.ycp
@@ -93,8 +93,8 @@ return ($[
 	    "pc104"	: $[ "ncurses": "ch-de_nodeadkeys.map.gz" ],
 	    "macintosh" : $[ "ncurses": "ch-de_mac.map.gz" ],
 	    "type4"	: $[ "ncurses": "us.map.gz" ],
-	    "type5"	: $[ "ncurses": "ch-de-sundeadkeys.map.gz" ],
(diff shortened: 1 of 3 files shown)


ancorgs:
> question Maybe we could do something similar here and regularly check that the defined keyboard maps are still available in the system.

Just to clarify, the keymaps are still available in all (open)SUSE systems. They were just moved to a `legacy` directory because they are generated with the old non-systemd way. So it's not exactly a case of maps disappearing little by little over time. It's more a consequence of TheNewWayOfDoingPrettyMuchEverything introduced by systemd.

I really hope we will rewrite the keyboard management of YaST soonish (see https://trello.com/c/dakSfQCr/3314-try-to-merge-the-rewritten-keyboard-management-in-yast2-country). With that, we will leave the gory details in the hands of systemd, instead of managing all kind of consolefonts & keyboard mapping directly in YaST.

PR #219, opened 2019-07-22
 Used deadkeys ("ch(basic)") for German (Switzerland) [boo#1142273] 

sndirsch:
Used deadkeys ("ch(basic)") for German (Switzerland) [boo#1142273]


sndirsch:
I need to redo this. :-(

PR #231, opened 2019-12-02
Master 1158119

schubi2:
## Problem

Keyboard has to be set in the proposal if AutoYaST has not already set it.

* Fast-track: https://trello.com/c/sL8vbVZI/1465-https-bugzillasusecom-showbugcgiid1158119

* This one is blocking Staging:Y from being accepted.

* https://openqa.suse.de/tests/3646052#step/installation/4

## Solution

Avoid to crash when no language and keyboard is defined.

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -116,6 +116,15 @@ module Yast
       @modified = true
     end
 
+    # Reset all setting
+    def reset
+      @curr_kbd = ""
+      @keyboard_on_entry = ""
+      @default_kbd = ""
+      @user_decision = false
+      @modified = false
+    end
+
     # Set current data into the installed system.
     def Save
       if Mode.update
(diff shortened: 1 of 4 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -316,6 +326,7 @@ def current_kbd
     publish :function => :Modified, :type => "boolean ()"
     publish :function => :SetModified, :type => "void (boolean)"
     publish :function => :Save, :type => "void ()"
+    publish :function => :reset, :type => "void ()"
```
joseivanlopez: Do we still need to publish methods?

keyboard/src/modules/Keyboard.rb
```diff
@@ -116,6 +116,15 @@ def SetModified
       @modified = true
     end
 
+    # Reset all setting
+    def reset
```
joseivanlopez: Where is this used? Please, document why it is needed.

schubi2: For the testcase only. Or do you see another chance to reset the current keyboard ?

joseivanlopez: What about defining private attr_readers and then mock it?

~~~
private

attr_reader :curr_kbd
attr_reader :keyboard_on_entry
attr_reader :default_kbd
attr_reader :user_decision
attr_reader :modified

# In the test

allow(subject).to receive(:curr_kbd).and_return("")
~~~

joseivanlopez: And make sure to use `!curr_kbd.empty?` instead of `!@curr_kbd.empty?` in the module code.

joseivanlopez:
> Please, try this https://github.com/yast/yast-country/pull/231/files#r352557334

Finally we decided to simply use a new instance in the test:

~~~
subject { Yast::KeyboardClass.new }
~~~

PR #232, opened 2019-12-04
Improve Windows detection

joseivanlopez:
### Problem

Detection of Windows system is quite inefficient.

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1135341
* Part of https://trello.com/c/vDzeJGkD/1413-follow-up-of-1135341-inefficient-check-for-installed-windows
* Requires https://github.com/yast/yast-storage-ng/pull/1001

### Solution

* Use new method provided by *yast2-storage-ng* to detect a Windows system.

### Tests

* Added unit tests

### DIFF
--- a/timezone/src/modules/Timezone.rb
+++ b/timezone/src/modules/Timezone.rb
@@ -1036,16 +1036,13 @@ module Yast
 
     # Checks whether the system has Windows installed
     def system_has_windows?
-      begin
-        win_partitions = disk_analyzer.windows_partitions
-        !win_partitions.empty?
-      rescue NameError => ex
-        # bsc#1058869: Don't enforce y2storage being available
-        log.warn("Caught #{ex}")
-        log.warn("No storage-ng support - not checking for a windows partition")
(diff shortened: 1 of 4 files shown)


timezone/src/modules/Timezone.rb
```diff
@@ -1036,16 +1036,13 @@ def Summary
...
-      rescue NameError => ex
-        # bsc#1058869: Don't enforce y2storage being available
-        log.warn("Caught #{ex}")
-        log.warn("No storage-ng support - not checking for a windows partition")
-        log.warn("Assuming UTC for the hardware clock")
-        false # No windows partition found
-      end
+      disk_analyzer.windows_system?
+    rescue NameError => ex
+      # bsc#1058869: Don't enforce y2storage being available
```
dgdavid: Just for confirmation, I guess this is the reason why you are not adding/updating a `yast-storage-ng` dependency in the .spec file, right?

joseivanlopez: Yes, it seems we don't want to have a direct dependency.

PR #233, opened 2019-12-18
 Setting keyboard: Ignoring empty keyboard values

schubi2:
While the installation process keyboard will be set at different places and different targets.
Depending the installation workflow the keyboard setting has already been evaluated/set or not.
It is "" if it has not been set. So in that case YaST will not try to write the keyboard setting now.

### DIFF
--- a/keyboard/test/kb_strategy_spec.rb
+++ b/keyboard/test/kb_strategy_spec.rb
@@ -58,7 +58,7 @@ describe Y2Keyboard::Strategies::KbStrategy do
     end
 
     context "empty keyboard_code parameter" do
-      it "returns only " do
+      it "does not try to set the keyboard layout" do
         expect(kb_strategy).not_to receive(:set_x11_layout)
         expect(Yast::Execute).not_to receive(:on_target!).with(
           "loadkeys", anything, anything)
--- a/keyboard/test/kb_strategy_spec.rb
+++ b/keyboard/test/kb_strategy_spec.rb
@@ -54,8 +54,16 @@ describe Y2Keyboard::Strategies::KbStrategy do
         expect(kb_strategy).to receive(:write_udev_rule)
 
         kb_strategy.set_layout("es")
-      end      
+      end
+    end
 
-    end    
+    context "empty keyboard_code parameter" do
+      it "returns only " do
+        expect(kb_strategy).not_to receive(:set_x11_layout)
+        expect(Yast::Execute).not_to receive(:on_target!).with(
+          "loadkeys", anything, anything)
+        kb_strategy.set_layout("")
+      end
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


keyboard/test/kb_strategy_spec.rb
```diff
@@ -54,8 +54,16 @@
         expect(kb_strategy).to receive(:write_udev_rule)
 
         kb_strategy.set_layout("es")
-      end      
+      end
+    end
 
-    end    
+    context "empty keyboard_code parameter" do
+      it "returns only " do
```
dgdavid: NP: "returns only" sounds strange. I'd say something like "does not try to set the keyboard layout".

keyboard/test/systemd_strategy_spec.rb
```diff
@@ -41,13 +41,23 @@
...
+          "localectl", "set-keymap", new_layout.code
+        )
 
-      systemd_strategy.apply_layout(new_layout.code)
+        systemd_strategy.apply_layout(new_layout.code)
+      end
+    end
+
+    context "empty keyboard code" do
+      it "returns only " do
```
dgdavid: As above, 

> NP: "returns only" sounds strange. I'd say something like "does not try to set the keyboard layout".

schubi2:
> Apart from the NPs, I have a question
> 
> Do you really think this fix https://bugzilla.suse.com/show_bug.cgi?id=1159210 or only _hides_ it? I mean, in the reported bug they are trying to set the keyboard with the "en" keyboard_code, which does not exist. But using the valid "us" or "uk"codes the problem is the same: [Keyboards.code("valid_keyboard_code")](https://github.com/yast/yast-country/blob/2b301ec423d986dcfc52a248d3755e8d55bc2b67/keyboard/src/modules/Keyboard.rb#L149) is going to return `nil` and, due the condition introduced here, the keyboard will be not set.
> 
> Just asking. I could try to reproduce it later if you want.
Good point I have fixed the regarding code because this error has happened  in the other bug too. But I will have to test this usecase too.

schubi2:
I have tested with the keyboard settings of the regarding bug too:
https://openqa.suse.de/tests/3700482/file/autoyast_profile.xml
<keyboard>
 <keymap>en</keymap>
</keyboard>
<language>
 <language>en_US</language>
 <languages>en_US</languages>
</language>
And it has worked :-)

dgdavid:
> I have tested with the keyboard settings of the regarding bug too:
https://openqa.suse.de/tests/3700482/file/autoyast_profile.xml
> 
> And it has worked :-)

As discussed in IRC, I think I didn't explain so well my concerns. I was trying to test it to show what I mean, but my virtual machine got stuck twice during the installation. So, let's not block this.

Thanks @schubi2

PR #234, opened 2019-12-19
Fix dialect

jreidinger:


language/src/clients/select_language.rb
```diff
@@ -591,11 +591,11 @@ def LanguageExpertDialog
       Builtins.y2debug("expert values %1", val)
 
       # get the list of locales for our language
-      lang = Builtins.substring(@language, 0, 2)
+      lang = @language[/^[a-z]+/]
```
ancorgs: What about extracting this to a method? Something like...

```ruby
def main_language(full_code)
  full_code[/^[a-z]+/]
end
```

Of course, with a better name. :-)

jreidinger: well, I am a bit scared by whole this module, but can do :)

PR #237, opened 2019-12-30
Fix loadkeys

jreidinger:
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1159185

Issue: string like "-C /dev/tty1 -C /dev/tty2" is passed to cheetah which expect one option per parameter. Result of keyboard refactoring.

Fix: pass properly each parameter per one argument.

### DIFF
--- a/keyboard/src/lib/y2keyboard/strategies/kb_strategy.rb
+++ b/keyboard/src/lib/y2keyboard/strategies/kb_strategy.rb
@@ -19,6 +19,7 @@
 
 require "yast"
 require "yast2/execute"
+require "shellwords"
 
 module Y2Keyboard
   module Strategies
@@ -31,7 +32,7 @@ module Y2Keyboard
     #
     # Use the systemd strategy for making keyboard changes permanent on
     # the installed system.
-    # 
+    #
     class KbStrategy
       include Yast::Logger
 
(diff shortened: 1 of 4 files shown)


keyboard/src/lib/y2keyboard/strategies/kb_strategy.rb
```diff
@@ -88,10 +89,10 @@ def set_x11_layout(keyboard_code)
       # It includes all tty[0-9]* and ttyS[0-9]* devices (bsc#1010938).
       #
       # @param [String] kind of tty ("tty", "ttyS")
-      # @return [String] ready to be passed to the loadkeys command
+      # @return [Array<String>] array with params for the loadkeys command
       def loadkeys_devices(kind)
         tty_dev_names = Dir["/dev/#{kind}[0-9]*"]
-        tty_dev_names.map { |d| "-C #{d.shellescape}" }.join(" ")
+        tty_dev_names.each_with_object([]) { |d,res| res << "-C" << d }
```
dgdavid: NP: Why not keep the interpolation?

```ruby
res << "-C #{d}"
```

dgdavid: Ah! forget it :)

PR #239, opened 2020-01-20
Keyboard definition in the AY configuration file: Alias AND keymap na…

schubi2:
…me is supported (bsc#1159541).
**Problem**
AY configuration file:
"Keymap" flag represents an keymap alias instead of the real keymap name.

**Solution**
Both are supported now.

### DIFF
--- a/keyboard/src/modules/Keyboard.rb
+++ b/keyboard/src/modules/Keyboard.rb
@@ -46,6 +46,7 @@ module Yast
       Yast.import "Mode"
       Yast.import "ProductFeatures"
       Yast.import "Stage"
+      Yast.import "Report"
 
       # general kb strategy which is used for temporary changes only.
       @kb_strategy = Y2Keyboard::Strategies::KbStrategy.new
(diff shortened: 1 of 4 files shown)


keyboard/src/modules/Keyboard.rb
```diff
@@ -281,6 +282,21 @@ def Import(settings, syntax = :keyboard)
...
+
+      # Checking if the keymap exists. Either it is the real keymap name
+      # or an alias.
+      if !Keyboards.code(keyboard)
+        # Checking if it a real keymap name
+        checked_keyboard = keyboard
+        keyboard = Keyboards.alias(checked_keyboard)
+        if !keyboard
+          # TRANSLATORS: the "%s" is the kaymap name
+          Report.Warning(_("Cannot find keymap: %s. Taking default one.") % checked_keyboard)
```
dgdavid: The question could be, what is the default one? We know it, of course. But, what about the user?

keyboard/src/modules/Keyboard.rb
```diff
@@ -281,6 +282,21 @@ def Import(settings, syntax = :keyboard)
       when :language
         keyboard = GetKeyboardForLanguage(settings["language"], keyboard)
       end
+
+      # Checking if the keymap exists. Either it is the real keymap name
+      # or an alias.
+      if !Keyboards.code(keyboard)
```
dgdavid: I would avoid this nested conditional. Something like 

```ruby
found_keyboard = Keyboards.code(keyboard) || Keyboards.alias(keyboard)

if !found_keyboard
  Report(...)
  return false
end
```

should work. What do you think?

schubi2: I do not see a real  benefit. I have these nested condition in order to assign the keyboard alias new and not only for displaying the error message.
At least I am out of ideas how it should be better ;-)

keyboard/src/modules/Keyboard.rb
schubi2: To be honest at that moment even we do not know it because the proposal will be called afterward. :-)

keyboard/src/modules/Keyboard.rb
dgdavid: Hmmm, even more, it could be an isolated method `find_keyboard`

```ruby
# @param keyboard [String] alias or keymap
def find_keyboard(keyboard)
  Keyboards.code(keyboard) || Keyboards.alias(keyboard)
end

...

if !find_keyboard
  ...
```

PR #243, opened 2020-02-19
[SLE-7188] Default to SUSE NTP servers

imobachgs:
Use the new `Y2Network::NtpServer` when proposing the default values for NTP servers.

Related PR: https://github.com/yast/yast-network/pull/1039

### DIFF
--- a/timezone/src/include/timezone/dialogs.rb
+++ b/timezone/src/include/timezone/dialogs.rb
@@ -32,6 +32,9 @@
 #	Dialogs for timeone and time configuration.
 #
 # $Id$
+
+require "y2network/ntp_server"
+
 module Yast
   module TimezoneDialogsInclude
     def initialize_timezone_dialogs(include_target)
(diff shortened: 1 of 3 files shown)


timezone/src/include/timezone/dialogs.rb
```diff
@@ -575,12 +577,7 @@ def TimezoneDialog(args)
         @ntp_used = true
         # configure NTP client
         # to prevent misusage of ntp.org we need to distinguish opensuse and SLE usage
-        base_products = Product.FindBaseProducts
-        if base_products.any? { |p| p["name"] =~ /openSUSE/i }
-          servers = (0..3).map { |i| "#{i}.opensuse.pool.ntp.org" }
-        else
-          servers = (0..3).map { |i| "#{i}.novell.pool.ntp.org" }
-        end
+        servers = Y2Network::NtpServer.default_servers.map(&:hostname)
```
teclator: It introduces a dependency on yast2-network >= 4.2.55 isn't it?

teclator: Maybe would be a good idea to conflict in a ntp-client version < 4.2.8 ?

https://github.com/yast/yast-ntp-client/pull/155/files#diff-2fd35b8b1087ae0be0605e2129e80172R20

imobachgs: Yes, it sounds like a good plan. Thanks!

PR #247, opened 2020-05-12
Do not try to restart non existing kbd service

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1170292

## Problem

after layout change a restart of kbd service is requested during write. As the service doesn't exist, everything is blocked for approx 1 minute due to a timeout

## Solution

Backported one line from https://github.com/yast/yast-country/pull/225

### DIFF
--- a/keyboard/src/clients/keyboard.rb
+++ b/keyboard/src/clients/keyboard.rb
@@ -127,7 +127,6 @@ module Yast
         Popup.ShowFeedback(Message.updating_configuration, Message.takes_a_while)
       end
       Keyboard.Save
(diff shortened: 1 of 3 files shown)


ancorgs:
Indeed, the `kbd` package do not longer includes such service, but it includes `kbdsettings.service` whose description is "Apply settings from /etc/sysconfig/keyboard".

I wonder whether the correct solution is to delete the service restarting (as this PR does) or changing it to use `kbdsettings` instead of `kbd`. Can you please check whether the former is a substitute of the latter?

PR #250, opened 2020-06-18
Generating AY file: Generate only values which are not default values.

schubi2:
Problem
========
Country settings will be exported although they are default values.

Solution
========
Comparing these values with the default values and do not export them if they are the same.


keyboard/src/lib/y2keyboard/clients/keyboard_auto.rb
```diff
@@ -37,7 +41,14 @@ def read
     end
     
     def export
-      Keyboard.Export
+      ret = Keyboard.Export
+      if ret["keymap"] == Keyboard.GetKeyboardForLanguage(Language.language,
```
imobachgs, line 45: Why is not `Keyboard.Export` responsible for this?

keyboard/src/lib/y2keyboard/clients/keyboard_auto.rb
```diff
@@ -37,7 +41,14 @@ def read
...
+           "english-us")
+        log.info("keymap #{ret["keymap"]} is the default of language"\
```
imobachgs, line 47: Please, add an space after "language".

language/src/lib/y2country/clients/language_auto.rb
```diff
@@ -0,0 +1,82 @@
...
+      {}      
+    end
+
+    def read
+      Language.Read(true)
+    end
+    
+    def export
+      ret = Language.Export
+      if Language.language == Language.default_language
```
imobachgs, line 52: Same tan above?

PR #251, opened 2020-06-23
Avoid yast2-country-data build dependency

joseivanlopez:
## Problem

Package yast2-country has a build dependency of its sub-package yast2-country-data. This is a chicken-egg problem. That build dependency was added here https://github.com/yast/yast-country/pull/244 because some issues with unit tests.

Now a build problem is raising when performing a full distro bootstrap.

* https://bugzilla.suse.com/show_bug.cgi?id=1173253

## Solution

Missing mocking was added to unit tests and now the build dependency is not needed anymore.

### DIFF
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jun 23 10:04:37 UTC 2020 - José Iván López González <jlopez@suse.com>
+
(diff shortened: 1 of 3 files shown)


joseivanlopez:
Test coverage has decreased significantly (-0.04%) because the mechanism to collect the timezones from the language flies is now mocked. IMHO, unit tests are missing for Language module.

PR #255, opened 2020-07-01
Expose default language

teclator:
Added a reader in order to check which is the current default language. (needed by yast/yast-firstboot#98)


teclator:
See https://github.com/yast/yast-firstboot/pull/98#discussion_r448445780

PR #259, opened 2020-10-12
adjust keyboard client layout to also fit the firstboot workflow (bsc#1174856)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1174856
- https://trello.com/c/mVt37N0M

The `firstboot_keyboard` client is out of date and doesn't work with the current `yast-country` code.

## Solution

I could not find a compelling reason why there has to be a separate `firstboot_keyboard` client in the first place. It seems pointless to more or less copy-and-paste reproduce the `keyboard` client code.

So why not use the `keyboard` client directly instead? All it needs (AFAICS) is adjusting the workflow buttons to be aligned with firstboot expectations.

## Screenshots

q
n

## See also

- https://github.com/yast/yast-firstboot/pull/107

### DIFF
--- a/keyboard/src/lib/y2keyboard/dialogs/layout_selector.rb
+++ b/keyboard/src/lib/y2keyboard/dialogs/layout_selector.rb
@@ -96,7 +96,9 @@ module Y2Keyboard
         finish_dialog(:abort)
       end
 
-      alias_method :abort_handler, :cancel_handler
+      def abort_handler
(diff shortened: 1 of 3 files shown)


shundhammer:
It makes a lot of sense to me, but I am not an expert here.

mvidner (approved):

PR #263, opened 2020-12-15
Removed old code for sysvinit configuration (bsc#1175494).

schubi2:


jreidinger (approved):

dgdavid:
Thanks @schubi2!

I thought that SysVinit dead code was only in yast2-security, see https://github.com/yast/yast-security/pull/79. Maybe you could also add a comment linking this PR in https://bugzilla.suse.com/show_bug.cgi?id=1175494

PR #266, opened 2021-03-11
Upstream/bsc1181952

jreidinger:
trello: https://trello.com/c/rQufN1iA/2333-sle-15-sp3-p2-1181952-keyboard-layout-is-wrong-for-some-languages
bzilla: https://bugzilla.suse.com/show_bug.cgi?id=1181952

Automated test that runs and verify codes are included.

PR #267, opened 2021-04-29
Fix failing unit test

teclator:
## Problem

As part of [this](https://github.com/yast/yast-country/pull/266) PR a unit test was introduced in order to verify the keyboard layout mapping.

https://github.com/yast/yast-country/pull/266/files#diff-299636c1b42b787c000e11ed239568da530b181d5c67886c9d2ea7f640896e26R11

The problem is that the the build package (which uses systemd-mini) does not contains the map generated by kbd.

So, the Jenkins jobs for submitting the changes failed because it expected /usr/share/systemd/kbd-model-map to contain 'es-ast', tw and cn.

## Solution

- Add systemd as a build dependency (maybe it is not the best solution but looks better than skip the test)

### DIFF
--- a/package/yast2-country.spec
+++ b/package/yast2-country.spec
@@ -15,7 +15,6 @@
 # Please submit bugfixes or comments via https://bugs.opensuse.org/
 #
 
-
 Name:           yast2-country
 Version:        4.3.14
 Release:        0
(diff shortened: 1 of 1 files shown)


dgdavid:
> This is not the best solution and requiring systemd for building the package is probably a bad idea. We could just skip the test by now at least in this branch.

Well, better to keep the test than skip it, isn't it?

dgdavid (approved):

teclator:
SR: https://build.suse.de/request/show/240391

PR #268, opened 2021-05-04
Merge SLE-15-SP3 into 'master'

teclator:
Fix failing build test in master too (merge #267)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on: [push, pull_request]
 jobs:
   Tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


lslezak:
I know this comes from an SP3 fix, but could we somehow mock the systemd call? To be honest, I do not like adding extra build dependency just because of an unit test, esp. if it is about systemd... :thinking:

ancorgs:
Mocking systemd would defeat the purpose of the test. It's not really a unit test. It's a test ensuring the layouts we have declared are actually present in a real SLE/Leap.

lslezak:
I see, then it makes sense...

lslezak (approved):

PR #274, opened 2021-07-22
Comment how to switch some keyboard layouts (bsc#1184452)

mvidner:
In https://bugzilla.suse.com/show_bug.cgi?id=1184452 testers say that some keyboard layouts don't work.
In fact they do work but the primary layout is US and you have to switch to the national layout. This is sometimes written as a comment in the .map.gz file, sometimes not even that.


ancorgs (approved):
LGTM. Thanks, this will help a lot for manual testing.

PR #273, opened 2021-06-30
Add normalized azerty layout

Cimbali:
This keyboard layout is the normalized (i.e. defined by the French authority of normalization AFNOR) version of the azerty layout. It’s a major improvement and was [released in 2019](https://normalisation.afnor.org/actualites/faq-clavier-francais/). It’s already made its way into `kbd` and `xkeyboard-config` a while ago:
```sh
> rpm -ql kbd | grep fr-afnor
/usr/share/kbd/keymaps/xkb/fr-afnor.map.gz
> grep -w afnor /usr/share/X11/xkb/symbols/fr
// Normalization: https://normalisation.afnor.org/actualites/faq-clavier-francais/
xkb_symbols "afnor" {
```

PR #275, opened 2021-07-29
Offer the fr-afnor keyboard only if it is defined (bsc#1188867) 

mvidner:
- https://bugzilla.suse.com/show_bug.cgi?id=1188867
- An amendment to #273: after merging it turned out that the internal Jenkins job fails because the SLE environment is different.

This change checks at run time whether the new French layout is present and filters it out if it's not.

There may be an option to add the layout to SLE15 instead, I'll ask the maintainers+release managers in Bugzilla.

### Testing

I have tested this by applying the following patch and inspecting the output of `rake osc:build` for `master` and `SLE-15-SP3`, with results being true and false respectively:
```diff
diff --git a/keyboard/test/data/keyboard_test.rb b/keyboard/test/data/keyboard_test.rb
index ed5c03a..10c1ccd 100644
--- a/keyboard/test/data/keyboard_test.rb
+++ b/keyboard/test/data/keyboard_test.rb
@@ -14,6 +14,8 @@ describe Keyboards do
         code = kb_map["code"]
         expect(valid_codes).to include(code)
       end
+      a = Keyboards.all_keyboards.any? { |k| k["code"] == "fr-afnor" }
+      puts "HAVE AFNOR: #{a}"
     end
   end
 end
```

### DIFF
--- a/keyboard/src/data/keyboards.rb
+++ b/keyboard/src/data/keyboards.rb
@@ -37,6 +37,11 @@ class Keyboards
   #   - suggested_for_lang [Array<String>] optional, language codes
   #       to suggest this layout for
   def self.all_keyboards
+    always_present_keyboards + optional_keyboards
+  end
+
+  # @see all_keyboards
+  def self.always_present_keyboards
     [
       { "description" => _("English (US)"),
         "alias" => "english-us",
@@ -64,11 +69,6 @@ class Keyboards
       { "description" => _("French"),
         "alias" => "french",
         "code" => "fr-latin1",
-        "suggested_for_lang" => ["br_FR", "fr", "fr_BE"]
-      },
-      { "description" => _("French (AFNOR)"),
-        "alias" => "french-afnor",
-        "code" => "fr-afnor",
         "suggested_for_lang" => ["br_FR", "fr", "fr_BE"]
       },
       { "description" => _("French (Switzerland)"),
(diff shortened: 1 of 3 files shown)


keyboard/src/data/keyboards.rb
```diff
@@ -266,6 +266,31 @@ def self.all_keyboards
...
+    end
+    afnor = {
+      "description" => _("French (AFNOR)"),
+      "alias" => "french-afnor",
+      "code" => "fr-afnor",
+      "suggested_for_lang" => ["br_FR", "fr", "fr_BE"]
+    }
+    @optional_keyboards.push(afnor) if afnor_test.call
+
+    @optional_keyboards
```
dgdavid, line 291: Maybe you can use `Yast::Execute` here

```suggestion
    # TODO: put below constants in a proper place
    GREP_CMD = "/usr/bin/grep".freeze
    KDB_MODEL_MAP = "/usr/share/systemd/kbd-model-map".freeze
    
    if Yast::Execute.on_target!(GREP_CMD, "^fr-afnor", KDB_MODEL_MAP)
      # The afnor layout was added to xkeyboard-config in 2019-06
      # but SLE15-SP4 only has 2.23 released in 2018
      @optional_keyboards.push(afnor_keyboard_layout)
    end
  rescue Cheetah::ExecutionFailed => e
    # TODO: log something?
  ensure
    @optional_keyboards
  end
  
  def afnor_keyboard_layout
    {
        "description" => _("French (AFNOR)"),
        "alias" => "french-afnor",
        "code" => "fr-afnor",
        "suggested_for_lang" => ["br_FR", "fr", "fr_BE"]
    }
  end
  
```

dgdavid: :thinking: :thinking: 

Still thinking that `Yast::Execute` could be an option, but not with the above approach since once it fails it will not add _the rest_ of optional keyboards when adding more.

mvidner: Well, "executing grep" is a nice conceptual shortcut that we're familiar with from the shell, and I 've thought of it too, but decided to implement it in Ruby directly. It is simple and faster.

joseivanlopez (approved):
LGTM. But there are conflicts in changelog.

joseivanlopez:
BTW, should we add an unit test?

mvidner:
> BTW, should we add an unit test?

No, we are fixing this because an *integration* test has failed. The test needs another package, can't *unit* test that.

PR #276, opened 2021-08-03
Do not write empty timezone

joseivanlopez:
## Problem

If the AutoYaST profile specifies an empty timezone (e.g., `<timezone/>`), then an error popup is shown at the end of the installation . Basically, the commands for setting the timezone (i.e., *systemd-firstboot --timezone* and *timedatectl set-timezone*) cannot be called without a timezone value.

* https://bugzilla.suse.com/show_bug.cgi?id=1188406

## Solution

Avoid to set the timezone if no timezone was given. Note that UTF timezone is used by default.

NOTE: The bug was reported for SLE-15-SP3, but it affects to SLE-15-SP2 as well and SP2 maintenance ends on 31 Dec 2021.

## Testing

* Unit tests.
* Manually tested.

### DIFF
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug  3 16:38:40 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Do not write an empty timezone (bsc#1188406).
+- 4.2.22
+
+-------------------------------------------------------------------
 Tue Jun 15 09:17:24 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Fix the Comment entry in the desktop file so the tooltip
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Tue Aug  3 16:38:40 UTC 2021 - José Iván López González <jlopez@suse.com>
 
-- Do not write an empty timezone (bsc#1188406).
+- AutoYaST: allow empty /profile/timezone/timezone setting,
+  meaning to keep the UTC default (bsc#1188406).
 - 4.2.22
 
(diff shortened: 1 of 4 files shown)


teclator (approved):
LGTM

teclator:
The bug report is against SLE-15-SP3, so, I guess the target SLE-15-SP2 is because it is inherited from there or have the same problem maybe would mention it in the PBI or in the bug.

package/yast2-country.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Aug  3 16:38:40 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Do not write an empty timezone (bsc#1188406).
```
mvidner, line 4: This is not helpful for the reader. I'll read the bug and suggest a better description

mvidner (changes requested):

package/yast2-country.changes
mvidner, line 4:
```suggestion
- AutoYaST: allow empty /profile/timezone/timezone setting,
  meaning to keep the UTC default (bsc#1188406).
```

joseivanlopez:
> The bug report is against SLE-15-SP3, so, I guess the target SLE-15-SP2 is because it is inherited from there or have the same problem maybe would mention it in the PBI or in the bug.

The package *yast2-country* is not inherited from SLE-15-SP2. But the bug affects to SLE-15-SP2 as well and SP2 maintenance ends on 31 Dec 2021.

mvidner:
Re bug validity:
AY docs does not mention this edge case: https://documentation.suse.com/sles/15-SP3/single-html/SLES-autoyast/#CreateProfile-Country
but the RNC schema says /profile/timezone/timezone is optional
So the user is reasonable in omitting it.

mvidner (approved):

joseivanlopez:
https://build.suse.de/request/show/247661

PR #279, opened 2021-08-18
Move keyboards database to make the module compatible with self-update

imobachgs:
Trello: https://trello.com/c/1g3hhqTA/
Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1189461

Trying to update yast2-country using the self-update mechanism causes a crash because `Keyboard.rb` module cannot load
the keyboards database (`keyboards.rb`) using a [require_relative](https://github.com/yast/yast-country/blob/5a212da88b5ca07431a10f8cde6d1be31eca7bd0/keyboard/src/modules/Keyboard.rb#L36) call.

The problem is that `keyboards.rb` and `Keyboard.rb` live in different packages (`yast2-country-data` and `yast2-country`), so on self-update they mounted as `/mounts/yast_00XX`, so `data/` and `modules/` are not side-by-side anymore.

I have decided to keep the changes to the minimum to make it work, but we should think about the way we ship the data files.


dgdavid (approved):
LGTM, thanks.

PR #280, opened 2021-08-19
Move keyboards database to make the module compatible with self-update (SLE-15-SP3)

imobachgs:
Merge #279 into `SLE-15-SP3`.


dgdavid (approved):
LGTM

PR #289, opened 2022-01-21
Unify YaST module mocking in unit tests

lslezak:
- Use the new YaST RSpec helper everywhere
- Additionally fixed changing the current language during test run which might result in failure like
```
         expected: (/selected language cannot be used/)
              got: ("Zvolený jazyk nelze používat ...
```
when the system has default Czech locale and the YaST translations are installed.
- Also fixed test failure in Tumbleweed where it reports `UTF-8` as the default encoding

## TODO

- [x] Inspect the failing Tumbleweed unit test


dgdavid (approved):
LGTM

PR #290, opened 2022-01-28
Add explicit glibc-locale dependency

dirkmueller:
The testsuite is using non-base locales so we need the full glibc-locale to pass it.


mvidner (approved):
Requiring the full locale package makes sense for a locale-dealing YaST module.
Thanks for the PR, Dirk!

mvidner:
@dirkmueller in what environment did you encounter this problem? OBS/IBS or manual testing?
I'll be making a Bugzilla entry for this, if you already don't have one.

PR #293, opened 2022-02-25
Fix typo portugese -> portuguese

elchevive:
Fix typo portugese -> portuguese ;-)

### DIFF
--- a/keyboard/src/lib/y2keyboard/keyboards.rb
+++ b/keyboard/src/lib/y2keyboard/keyboards.rb
@@ -138,17 +138,17 @@ class Keyboards
         "suggested_for_lang" => ["fa_IR"]
       },
       { "description" => _("Portuguese"),
-        "alias" => "portugese",
+        "alias" => "portuguese",
         "code" => "pt",
         "legacy_code" => "pt-latin1"
       },
       { "description" => _("Portuguese (Brazil)"),
-        "alias" => "portugese-br",
+        "alias" => "portuguese-br",
         "code" => "br",
         "legacy_code" => "br-abnt2"
       },
       { "description" => _("Portuguese (Brazil -- US accents)"),
-        "alias" => "portugese-br-usa",
+        "alias" => "portuguese-br-usa",
         "code" => "br-nativo-us",
         "legacy_code" => "us-acentos"
       },


ancorgs:
First of all, thanks for your contribution.

But just for the record - that's an internal identifier that, as far as I know, is never displayed in the user interface and has no influence in which exact keymap is applied (that would be the "code" field).

Or am I wrong and this is visible somehow?

On the other hand, I wonder if changing this could introduce any incompatibility. The first that comes to my mind would be when using YaST in the command line (not to be confused with the ncurses interface). For example, this command would stop working (it would be replaced by a more correct version... but backwards incompatible).

```
yast2 keyboard set portugese
```

So this needs some analysis in order to check what can break (specially since it seems to not fix any visible issue).

shundhammer:
It's also wrong in the user-visible texts AFAICS.

shundhammer:
Those identifiers that you (absolutely correctly) identified as misspelled are used for AutoYaST, and they have been in use since forever: Admins using AutoYaST are probably aware of the misspelling, and I fear they used that misspelling in their existing AutoYaST XML profiles.

So while it would certainly be nice to fix that misspelling, I fear we might break more than we fix; it would mean becoming incompatible with those existing AutoYaST profiles.

shundhammer:
I just checked, and it goes back to an ancient file `keyboard_raw.ycp` which must go back to around 2002 or so. And I bet even the user description has been misspelled all that time as well.

elchevive:
I found this while reading one of your summary posts, feel free to discard this change if its too much work for too little benefit ;-)

shundhammer:
Don't get us wrong: Your contribution is definitely welcome. And IMHO we will want to fix the user-visible messages (the ones wrapped in `_("Portuguese")`. I think we should be conservative with the `alias` field, though, and rather remain bug-compatible.

We just have to determine how far the translation process for the upcoming SLE-15 SP4 / Leap 15.4 release is because otherwise the translations will break, so every user would see the (now correctly spelled) English "Portugese".

ancorgs:
The visible texts have been using the correctly spelled "Portuguese" for years. This pull request only fixes the wrongly spelled internal identifiers.

As explained, fixing that would risk the compatibility in the command line and AutoYaST. So I guess is safer to keep the historical identifier despite the error.

Thanks for the attempt anyway. It was a nice one.

PR #298, opened 2022-10-12
switch to 'multix' variant for Candian CSA layout

sndirsch:
Old 'multi' variant for 'ca' keyboard layout has been deleted upstream with xkeyboard-config 2.37, but newer and less complex 'multix' stays. So switch to the latter.
https://github.com/freedesktop/xkeyboard-config/commit/d46d0f40afb8cb7842959bdd15dfccbc9c041493


jreidinger (approved):

mvidner:
@sndirsch Thanks, Stefan!

PR #302, opened 2023-01-10
Stop using File.exists? which no longer works in Ruby 3.2 (bsc#1206419)

mvidner:
## Problem

Ruby 3.2 removed `File.exists?` which has been long deprecated in favor of `File.exist?`

- https://trello.com/c/jRe5bWVE
- https://bugzilla.suse.com/show_bug.cgi?id=1206419
- build monitor: https://build.opensuse.org/project/monitor/openSUSE:Factory:Staging:H?arch_x86_64=1&defaults=0&failed=1&repo_standard=1

## Solution

The solution is a one-character diff.

## Testing

It fails during unit testing already, good.

## Screenshots

N/A

PR #303, opened 2023-01-10
Fix build with Ruby 3.2

DimStar77:
File.exists has been deprecated since Ruby 2.1

## Problem

Code fails to build with Ruby 3.2 due to usage of deprecated File.exists

- https://bugzilla.opensuse.org/show_bug.cgi?id=1206419

## Solution

Use File.exist, which was introduced in Ruby 2.1


DimStar77:
ouch.. closing in favor of https://github.com/yast/yast-country/pull/302 - Missed that PR :)

PR #304, opened 2023-01-11
Read local timezone database

joseivanlopez:
## Problem

Regions and time zones are empty when running "Date and Time" client from a YaST container in ALP.

## Solution

The list of regions is empty because [the database is being read from the host system](https://github.com/yast/yast-country/blob/master/timezone/src/modules/Timezone.rb#L226) (ALP), but  the database is in the container file system. @jreidinger suggested to use WFM instead of SCR to read the database from the local file system (i.e., the container). Doing that, the UI is now correctly populated.

Note: there is still an issue when executing `timedatectl` in host for saving the changes. This will be addressed in a separate bug because is a more generic issue related to SELinux and D-Bus.

## Testing

* Tested manually in running system, installer and YaST container.

### DIFF
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jan 11 12:13:52 UTC 2023 - José Iván López González <jlopez@suse.com>
+
+- Always read local timezone database (bsc#1207000).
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

PR #307, opened 2023-04-20
Cleanup: use "ru" keymap for Russian, not "ruwin_alt-UTF-8"   (bsc#1194609)

mvidner:
## Problem

The SUSE systemd maintainer wants to upstream a patch to `kbd-model-map`

- https://build.opensuse.org/package/view_file/Base:System/systemd/kbd-model-map.legacy?expand=1
- related: https://bugzilla.suse.com/show_bug.cgi?id=1194609

## Solution

Cleanup: use "ru" keymap for Russian, not "ruwin_alt-UTF-8"

ruwin_alt-UTF-8 was introduced in 2008 in response to https://bugzilla.suse.com/show_bug.cgi?id=432862 because the previous keymap did not use Unicode.

"ru" does use Unicode, and it has a dual latin-cyrillic layout (AltGr for single characters, Ctrl-Shift for layout switching)

In /usr/share/systemd/kbd-model-map the Russian keymap is a kbd-model-map.legacy SUSE patch over upstream, and as Franck Bui was reviewing this patch for upstreaming, we found that an entry for Russian is still needed, but a default "ru" keymap has a better chance than a specific one.

## Testing

- we already have a "unit" test checking that the YaST layout is is present in `/usr/share/systemd/kbd-model-map`

## Screenshots

No

### DIFF
--- a/dropping_kbd_legacy.md
+++ b/dropping_kbd_legacy.md
@@ -67,7 +67,7 @@ once, so removing kbd-legacy would break these languages als in X11.
 | Legacy keyboard map  | Selected replacement | Other options | Note           |
 | -------------------- | -------------------- | ------------- | ---------      |
 | gr                   | ?                    |               | Greek          |
-| ruwin_alt-UTF-8      | ?                    |               | Russian        |
+| ru                   | ?                    |               | Russian        |
 | tj_alt-UTF8          | ?                    |               | Tajik          |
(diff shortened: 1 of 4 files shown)


jreidinger (approved):

fbuihuu:
May I ask whether this change will reach next version of SLE 15 (ie SP6) ?

PR #309, opened 2023-05-10
Use localectl instead of reading kbd-model-map (bsc#1211104)

mvidner:
## Problem

- [bsc#1211104](https://bugzilla.suse.com/show_bug.cgi?id=1211104)

> ### yast-country: please stop parsing /usr/share/systemd/kbd-model-map in the unit tests

>  Basically systemd doesn't need to rely on /usr/share/systemd/kbd-model-map.xkb-generated to convert generated keymaps (those located in /usr/share/kbd/keymaps/xkb/) into x11 layouts, this is done automatically.
> 
> Hence /usr/share/systemd/kbd-model-map has been drastically shrunk and all entries that were previously taken from kbd-model-map.xkb-generated were dropped.
> 
> But it chagrins one of the yast-country unit tests
> 

## Solution

- at run time, call `localectl` instead
- remove the build time test
    - to be replaced by an openQA integration test

## Testing

- *Tested manually*
- and realized the UI does not sort the selection box! And fixed that too
- [ ] test building in the staging project with the new systemd, before and after

## Screenshots

Sorted now, respecting the locale

yast-country-sorted-keyboards

### DIFF
--- a/keyboard/src/lib/y2keyboard/keyboards.rb
+++ b/keyboard/src/lib/y2keyboard/keyboards.rb
@@ -32,8 +32,8 @@ class Keyboards
   #   - alias [String] yast-internal keybord id, to match the "keyboard" key
   #       in language/src/data/languages/language_*.ycp
   #   - code [String] keyboard name used by kbd, and
-  #       present in /usr/share/systemd/kbd-model-map
-  #       (test/data/keyboard_test.rb checks this)
+  #       present in `localectl list-keymaps`
+  #       (FIXME: t/... checks this)
   #   - legacy_code [String] old keyboard name used by kbd-legacy,
   #       present here so it can be automatically replaced if found
   #       in existing configurations (upgrade and AutoYaST profiles);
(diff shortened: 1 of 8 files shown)


keyboard/src/lib/y2keyboard/keyboards.rb
```diff
@@ -340,24 +340,23 @@ def self.optional_keyboards
     # memoize this
     return @optional_keyboards unless @optional_keyboards.nil?
 
-    @optional_keyboards = []
+    localectl_keymaps = `localectl list-keymaps`.split("\n")
```
shundhammer, line 343: Can we rely on `localectl` always being available on all products?

I know it's part of the _systemd_ package, at least on my Leap 15.4. Can we / do we want to rely on that?

Or maybe check here if the command is actually there?

shundhammer (approved):
LGTM with the caveat of `localectl` possibly not always being available (see my other comment).

dgdavid:
As commented in the IRC channel, according to the [comment #4](https://bugzilla.suse.com/show_bug.cgi?id=1211104#c4) in the bug report, keeping this PR open does not have too much sense.

> It turns out that also yast2-x11 contains a custom tool xkbctrl that also uses kbd-model-map.
https://github.com/search?q=org%3Ayast%20xkbctrl&type=code
> 
> In some contexts it can be replaced by localectl but it wouldn't work at installation time.

Thanks @mvidner!

PR #315, opened 2023-10-06
use also kbd-model-map.xkb-generated (bsc#1211104)

mvidner:
## Problem

(same problem as in #309)

- [bsc#1211104](https://bugzilla.suse.com/show_bug.cgi?id=1211104)

> ### yast-country: please stop parsing /usr/share/systemd/kbd-model-map in the unit tests

>  Basically systemd doesn't need to rely on /usr/share/systemd/kbd-model-map.xkb-generated to convert generated keymaps (those located in /usr/share/kbd/keymaps/xkb/) into x11 layouts, this is done automatically.
> 
> Hence /usr/share/systemd/kbd-model-map has been drastically shrunk and all entries that were previously taken from kbd-model-map.xkb-generated were dropped.
> 
> But it chagrins one of the yast-country unit tests
> 

## Solution

(The previous attempt to call `localectl` did not work in inst-sys)

- read both `kbd-model-map` and `kbd-model-map.xkb-generated` here
- also in `xkbctrl` in yast2-country: https://github.com/yast/yast-x11/pull/27

## Testing

- Tested manually with `rake run[keyboard]`
- and realized the UI does not sort the selection box! And fixed that too

## Screenshots

Sorted now, respecting the locale

yast-country-sorted-keyboards

### DIFF
--- a/keyboard/src/lib/y2keyboard/keyboards.rb
+++ b/keyboard/src/lib/y2keyboard/keyboards.rb
@@ -334,6 +334,21 @@ class Keyboards
     ]
   end
 
+  # @return [Array<String>]
+  def self.kbd_model_map_lines
+    filenames = [
+      "/usr/share/systemd/kbd-model-map",
+      # https://bugzilla.suse.com/show_bug.cgi?id=1211104
+      "/usr/share/systemd/kbd-model-map.xkb-generated"
+    ]
+    lines = filenames.map do |fn|
+      File.readlines(fn)
+    rescue Errno::ENOENT
+      []
+    end.flatten
+    lines
+  end
+
   # Some keyboards are present in new openSUSE releases but not in older SLE
   # @see all_keyboards
   def self.optional_keyboards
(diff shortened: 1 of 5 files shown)


joseivanlopez (approved):
LGTM

keyboard/src/lib/y2keyboard/keyboards.rb
```diff
@@ -334,6 +334,21 @@ def self.always_present_keyboards
...
 
+  # @return [Array<String>]
+  def self.kbd_model_map_lines
+    filenames = [
+      "/usr/share/systemd/kbd-model-map",
+      # https://bugzilla.suse.com/show_bug.cgi?id=1211104
+      "/usr/share/systemd/kbd-model-map.xkb-generated"
+    ]
+    lines = filenames.map do |fn|
+      File.readlines(fn)
```
lslezak, line 345: IMHO exceptions should used only for *real* exceptions, if a file is optional I'd prefer using `File.exist` instead.

The code is then even simpler:
```ruby
File.exist(fn) ? File.readlines(fn) : []
```

lslezak (commented):
No exceptions?

PR #323, opened 2024-05-20
Rename Europe/Kiev to Europe/Kyiv to align with timezone 2022b onwards

Dzejrou:
## Problem

In August 2022, the release of tz data/code of version 2022b renamed the Europe/Kiev timezone to Europe/Kyiv:

https://mm.icann.org/pipermail/tz-announce/2022-August/000071.html

This PR fixes the occurrences of the timezone name in yast-country.

This change has a caveat: With yast-translations not being updated, the new timezone label will not match any msgid and thus will not be translated. For example, I set my system to Czech localization and executed `yast2 timezone`, where I saw:

> Turecko
> Ukraine (Kyiv) <-- not translated
> Ukrajina (Simferopol) 

I'm planning to open also a PR for yast-translations that changes all msgid from 'Ukraine (Kiev)' to 'Ukraine (Kyiv)' with the msgstr being untouched (the translations are still the same) because the timezone 'Ukraine (Kiev)' no longer exists and that msgid (at least to my knowledge) has no other use.

- https://bugzilla.suse.com/show_bug.cgi?id=1224387 (L3)

### DIFF
--- a/language/src/data/languages/language_uk_UA.ycp
+++ b/language/src/data/languages/language_uk_UA.ycp
@@ -36,7 +36,7 @@
 		    _("Ukrainian")
 	],
 	// 2. what time zone propose for this language
-	"timezone"	: "Europe/Kiev",
+	"timezone"	: "Europe/Kyiv",
 	// 3. which keyboard layout propose for this language
 	"keyboard"	: "ukrainian",
     ];
--- a/package/yast2-country.changes
+++ b/package/yast2-country.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu May 16 20:41:25 UTC 2024 - jjindrak@suse.com
+
+- Rename Europe/Kiev to Europe/Kyiv as per 2022b release of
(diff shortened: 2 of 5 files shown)


shundhammer:
Please rebase this to the SLE-15-SP5 branch because that's where this patch is wanted in that L3 bug.

Our procedure is to start at the earliest affected branch, merge the PR, then merge that branch to the next newer branch (SLE-15-SP6 in this case), merge the PR, then merge that branch (SLE-15-SP6) to the next newer (master in this case) and merge again.

That way we cannot lose patches from earlier branches, and we can always simply merge upwards without conflicts, and the Git parent SHA1s remain the same; which all we would not get with cherry-picking individual patches from one branch to the next.

Dzejrou:
> Please rebase this to the SLE-15-SP5 branch because that's where this patch is wanted in that L3 bug.

Opened https://github.com/yast/yast-country/pull/324 for SLE-15-SP5, so I'll close this PR.

PR #324, opened 2024-05-28
Rename Europe/Kiev to Europe/Kyiv to align with timezone 2022b onwards

Dzejrou:
## Problem

In August 2022, the release of tz data/code of version 2022b renamed the Europe/Kiev timezone to Europe/Kyiv:

https://mm.icann.org/pipermail/tz-announce/2022-August/000071.html

This PR fixes the occurrences of the timezone name in yast-country.

This change has a caveat: With yast-translations not being updated, the new timezone label will not match any msgid and thus will not be translated. For example, I set my system to Czech localization and executed yast2 timezone, where I saw:

    Turecko
    Ukraine (Kyiv) <-- not translated
    Ukrajina (Simferopol)

I'm planning to open also a PR for yast-translations that changes all msgid from 'Ukraine (Kiev)' to 'Ukraine (Kyiv)' with the msgstr being untouched (the translations are still the same) because the timezone 'Ukraine (Kiev)' no longer exists and that msgid (at least to my knowledge) has no other use.

*    https://bugzilla.suse.com/show_bug.cgi?id=1224387 (L3)

PR #325, opened 2024-05-29
Rename Europe/Kiev to Europe/Kyiv [SLE-15-SP6]

shundhammer:
## Target Branch

This is a merge of #324 to SLE-15-SP6, to be merged later to _master_ as well.

## Problem

The _Europe/Kiev_ timezone was renamed to _Europe/Kyiv_. See #324 for details.

## Merging

This will be on hold until our CI is set up to target `SLE-15-SP6:Update`. This will take a few days.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,10 @@ jobs:
     - name: Git Checkout
       uses: actions/checkout@v2
 
+    - name: Install Needed Packages
+      # kbd is needed for runing the unit tests
+      run:  zypper --non-interactive install --no-recommends kbd
+
     # just for easier debugging...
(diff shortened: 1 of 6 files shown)


shundhammer:
Have the keyboards changed in SLE-15-SP6? The unit tests are now failing.

lslezak:
Fixed CI by installing the `kbd` package.

The package is quite big so it is installed only here and not in the shared container images because this is the only package which needs it.

```
The following 2 NEW packages are going to be installed:
  kbd kbd-legacy

2 new packages to install.
Overall download size: 2.3 MiB. Already cached: 0 B. After the operation, additional 4.5 MiB will be used.

```

lslezak (approved):

PR #329, opened 2025-05-21
Load required "setxkbmap" package on demand (bsc#1243088)

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1243088

## Problem

The `yast2 keyboard` module uses the external `setxkbmap` command from the _setxkbmap_ package to change the keyboard layout on the fly so the user can test it, but it didn't ensure that package was even installed, so the user got an error.

It was even more confusing because that error also appeared when the user closed the dialog with the `[x]` window manager close button at the top right corner.

## Fix 

Now checking if we are running in graphical mode (in the Qt UI), and if yes, installing the `setxkbmap` package after prompting the user.

## Why the error when the WM_CLOSE button was used?

The keyboard is changed on the fly immediately when clicking in the list of available layouts, and if the user cancels the dialog (with the "Cancel" button or the `[x]` WM_CLOSE button), the previous keyboard layout is restored, which uses the same command.

Otherwise the user would be stuck with whatever keyboard layout was last selected from the list for the duration of the current graphical session.

### DIFF
--- a/keyboard/src/lib/y2keyboard/clients/keyboard.rb
+++ b/keyboard/src/lib/y2keyboard/clients/keyboard.rb
@@ -110,14 +110,29 @@ module Yast
       end
     end
 
+    # Make sure that all needed packages are installed
+    #
+    # @return [Boolean]
+    def self.required_packages_installed
+      return true if UI.TextMode
+
+      # setxkbmap is needed in dialogs/layout_selector.rb for changing the X11
+      # keyboard layout on the fly, and for restoring the original one if the
+      # user cancels the dialog (bsc#1243088).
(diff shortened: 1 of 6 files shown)


shundhammer:
This is now **only the version bump and change log entry**, and it still fails in GitHub CI.

So it wasn't any of my changes that broke it, it was broken all along.

shundhammer:
The _relevant_ change would have been 7fa02482204b830b1f0a439891c1aced31b306da.

shundhammer:
Now some unit tests again fail _only in GitHub CI_. :disappointed:

joseivanlopez (approved):
LGTM

